PR #5, opened 2020-04-23
First pass adding samples folders

disa6302:
*Issue #, if available:*

*Description of changes:*

- Added webrtc and producer c samples folders


Sean-Der:
I would put a `CmakeLists.txt` in each sample directory and have that download dependencies. If we do submodules the issue is that people can't copy this into their own projects. 

I would do the directory structure like this, since we have some samples that use WebRTC and Producer

* `webrtc-with-gstreamer`
* `producer-with-gstreamer`
* `webrtc-and-producer-from-file`

etc... then at anytime someone copy that directory out and it would continue to work. Please push back if you disagree!

PR #8, opened 2020-06-03
Initialize project structure

lherman-cs:
The main goal for the structure is to give easy access/navigation to KVS’s common use cases for customers, similar to how Chime does it for their [JS SDK](https://github.com/aws/amazon-chime-sdk-js) but for multiple libraries. 

The structure is meant to be use case oriented. Each use case folder must have some descriptions about the use case. A use case folder might not contain sample codes, but only high level diagrams/explanations.

If a use case folder contains sample codes, the README should have the steps how to setup the project. The sample codes should not rely on other sample codes from other use cases. This also applies to the build systems. Consequently, each use case is **self isolated**.

By self isolating each use case, it'll reduce customer burden, customers should be able to use our samples to integrate our libraries easily to their projects and focus on a use case folder.

Format:

```
* README.md (list of KVS libraries and their use cases)
* <library 1>
    * <use case 1>
        * README.md (descriptions about the use case and how to setup/run or high level diagrams/concept)
    * <use case 2>
    * <use case n>
* <library 2>
* <library n>
```


disa6302:
Looks good. How do we plan on setting up the build system for this?
Since we are targeting demos from different assets.

lherman-cs:
I think we should have a minimal build system for each use case. It'll definitely put extra work on our side, but I think it has positive effect to the customers. They should be able to integrate our samples to their projects and navigate the samples easier.

disa6302:
Agreed. We can have CMake files to pull in each SDK repo based on the samples we are running right? And then we would build the requisite dependencies by executing the make file, right? Is that your thought?

lherman-cs:
We can simply use either [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) or [FetchContent](https://cmake.org/cmake/help/v3.11/module/FetchContent.html) from Cmake.

PR #15, opened 2020-07-14
End to end latency calc bug

disa6302:
*Issue #, if available:*

*Description of changes:*
A bug was introduced in calculating the end to end latency wherein negative values are observed if putUnalignedInt version calls are used. The PR includes a fix for this and few other cleanup

### DIFF
--- a/producer-c/producer-cloudwatch-integ/KvsProducerSampleCloudwatch.cpp
+++ b/producer-c/producer-cloudwatch-integ/KvsProducerSampleCloudwatch.cpp
@@ -11,14 +11,13 @@ VOID sigintHandler(INT32 sigNum)
 // add frame pts, frame index, original frame size, CRC to beginning of buffer
 VOID addCanaryMetadataToFrameData(PFrame pFrame) {
     PBYTE pCurPtr = pFrame->frameData;
-    putUnalignedInt64(pCurPtr, pFrame->presentationTs / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    putUnalignedInt64BigEndian((PINT64)pCurPtr, pFrame->presentationTs / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
     pCurPtr += SIZEOF(UINT64);
-    putUnalignedInt32(pCurPtr, pFrame->index);
+    putUnalignedInt32BigEndian((PINT32)pCurPtr, pFrame->index);
     pCurPtr += SIZEOF(UINT32);
-    putUnalignedInt32(pCurPtr, pFrame->size);
+    putUnalignedInt32BigEndian((PINT32)pCurPtr, pFrame->size);
     pCurPtr += SIZEOF(UINT32);
-    MEMSET(pCurPtr, 0x00, SIZEOF(UINT64));
-    putUnalignedInt64(pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
+    putUnalignedInt32BigEndian((PINT32)pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
 }
 
 VOID createCanaryFrameData(PFrame pFrame) {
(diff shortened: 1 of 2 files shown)


producer-c/producer-cloudwatch-integ/KvsProducerSampleCloudwatch.cpp
```diff
@@ -11,14 +11,14 @@ VOID sigintHandler(INT32 sigNum)
 // add frame pts, frame index, original frame size, CRC to beginning of buffer
 VOID addCanaryMetadataToFrameData(PFrame pFrame) {
     PBYTE pCurPtr = pFrame->frameData;
-    putUnalignedInt64(pCurPtr, pFrame->presentationTs / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    putInt64((PINT64)pCurPtr, pFrame->presentationTs / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
```
MushMal, line 14: Not sure how this fixes bad values

producer-c/producer-cloudwatch-integ/KvsProducerSampleCloudwatch.cpp
```diff
@@ -11,14 +11,14 @@ VOID sigintHandler(INT32 sigNum)
...
     pCurPtr += SIZEOF(UINT64);
-    putUnalignedInt32(pCurPtr, pFrame->index);
+    putInt32((PINT32)pCurPtr, pFrame->index);
     pCurPtr += SIZEOF(UINT32);
-    putUnalignedInt32(pCurPtr, pFrame->size);
+    putInt32((PINT32)pCurPtr, pFrame->size);
     pCurPtr += SIZEOF(UINT32);
     MEMSET(pCurPtr, 0x00, SIZEOF(UINT64));
-    putUnalignedInt64(pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
+    putInt32((PINT32)pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
```
MushMal, line 21: Previously , it was uint64 which was wrong

producer-c/producer-cloudwatch-integ/KvsProducerSampleCloudwatch.cpp
```diff
@@ -11,14 +11,14 @@ VOID sigintHandler(INT32 sigNum)
...
     pCurPtr += SIZEOF(UINT64);
-    putUnalignedInt32(pCurPtr, pFrame->index);
+    putUnalignedInt32((PINT32)pCurPtr, pFrame->index);
     pCurPtr += SIZEOF(UINT32);
-    putUnalignedInt32(pCurPtr, pFrame->size);
+    putUnalignedInt64((PINT32)pCurPtr, pFrame->size);
     pCurPtr += SIZEOF(UINT32);
-    MEMSET(pCurPtr, 0x00, SIZEOF(UINT64));
-    putUnalignedInt64(pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
+    MEMSET(pCurPtr, 0x00, SIZEOF(UINT32));
```
MushMal, line 20: Sorry, one more thing.. We don't really need the memset right? The next line will actually overwrite pCurPtr with the checksum anyway. I am not sure why this was there in the first place

disa6302: You are right. I had added it at the time for accuracy. 
Also, it looks like putUnalignedInt64 leads to a different value and putInt64 leads to a different value.

disa6302: For some reason, the end to end latency leads to very large values (negative) when I use putUnalignedInt64 but works fine when I use putInt64.

MushMal: There shouldn't be any difference. If we don't get a bus error (on platforms that don't support alignment) then we should be fine. There is a potential for endianness mismatch perhaps. Can we debug into this? This is a straight forward issue really

PR #26, opened 2020-08-11
Init WebRTC Canary

lherman-cs:
## Installation

```sh
cmake ..
```

## Build

```sh
make kvsWebrtcCanary
```

## Usage

```
CANARY_USE_TURN=ON CANARY_TRICKLE_ICE=ON CANARY_CLIENT_ID="MASTER" CANARY_IS_MASTER=TRUE CANARY_CHANNEL_NAME="ScaryTestChannel" CANARY_LOG_GROUP_NAME="canary" CANARY_CERT_PATH=../certs/cert.pem ./kvsWebrtcCanary
```

## TODO:
- Make viewer sequence works
- Add metrics
- Renaming
- Refactor/clean up


disa6302:
My only concern about a complete C++ solution is if customers would like to run canaries of their own, they will have to modify their files to a great degree to adapt to a C++ solution, which might be a lot of changes on their end.

webrtc-c/canary/src/Include.h
```diff
@@ -0,0 +1,51 @@
...
+
+#define NUMBER_OF_H264_FRAME_FILES  1500
+#define NUMBER_OF_OPUS_FRAME_FILES  618
+#define SAMPLE_VIDEO_FRAME_DURATION (HUNDREDS_OF_NANOS_IN_A_SECOND / DEFAULT_FPS_VALUE)
+#define SAMPLE_AUDIO_FRAME_DURATION (20 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+
+#define ASYNC_ICE_CONFIG_INFO_WAIT_TIMEOUT (3 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+#define ICE_CONFIG_INFO_POLL_PERIOD        (20 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+
+#define CANARY_CHANNEL_NAME_ENV_VAR    "CANARY_CHANNEL_NAME"
```
disa6302, line 21: One thing I have in mind: Should we create a shell script with exports of all this so that all the customer needs to do is modify values in the script and run it? 

I feel it will be a lot more cleaner and easier to keep track of while running the app.

webrtc-c/canary/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+  src/Main.cpp)
+target_link_libraries(
+  kvsWebrtcCanary
+  kvsWebrtcClient
+  kvsWebrtcSignalingClient
+  kvspicUtils
+  aws-cpp-sdk-core
+  aws-cpp-sdk-monitoring
+  aws-cpp-sdk-logs)
+
```
disa6302, line 50: I think we should set CERT_PATH to point to cert.pem by default like we do in the webrtc repo. cert path need not be configurable as part of canary

webrtc-c/canary/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+project(KVSWebRTCCanary LANGUAGES C CXX)
+
+set(CMAKE_CXX_STANDARD 11)
+
+include(FetchContent)
+
+# AWS SDK options
+set(CUSTOM_MEMORY_MANAGEMENT OFF)
+set(ENABLE_TESTING OFF)
+set(BUILD_ONLY "monitoring;logs")
```
disa6302, line 11: I think we should build events as well like we do in producer C. From what I recall, we do use it somewhere in producer C atleast.

webrtc-c/canary/src/CloudwatchLogs.cpp
```diff
@@ -0,0 +1,109 @@
...
+    this->client.CreateLogGroup(createLogGroupRequest);
+
+    createLogStreamRequest.SetLogGroupName(pConfig->pLogGroupName);
+    createLogStreamRequest.SetLogStreamName(pConfig->pLogStreamName);
+    createLogStreamOutcome = this->client.CreateLogStream(createLogStreamRequest);
+
+    if (!createLogStreamOutcome.IsSuccess()) {
+        CHK_ERR(FALSE, STATUS_INTERNAL_ERROR, "Failed to create \"%s\" log stream: %s", pConfig->pLogStreamName,
+                createLogStreamOutcome.GetError().GetMessage().c_str());
+    }
```
MushMal, line 28: You could do CHK_ERR(!createLogStreamOutcome.IsSuccess(), STATUS_OPERATION_FAILED, "blah blah);

webrtc-c/canary/src/Cloudwatch.h
```diff
@@ -0,0 +1,63 @@
...
+
+  private:
+    PConfig pConfig;
+    CloudWatchLogsClient client;
+    Synchronization sync;
+    Aws::Vector<InputLogEvent> logs;
+    Aws::String token;
+};
+
+class CloudwatchMonitoring {
```
MushMal, line 28: Perhaps split this into it's own header like the source?

webrtc-c/canary/src/Cloudwatch.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+namespace Canary {
+
+class Synchronization {
```
MushMal, line 5: Will this be only used here? If only used in CW logs then should it be encapsulated within that class as private?

webrtc-c/canary/src/Config.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    return retStatus;
+}
+
+BOOL getenvBool(CHAR const* pKey)
+{
+    PCHAR pValue;
+
+    pValue = getenv(pKey);
+    if (pValue == NULL) {
```
MushMal, line 23: Why not use mustenv here

webrtc-c/canary/src/Config.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+CleanUp:
+
+    return retStatus;
+}
+
+BOOL getenvBool(CHAR const* pKey)
+{
+    PCHAR pValue;
+
```
MushMal, line 21: Really minor: I recommend people using a single return from within a function. For a super simple one like this it's OK. However, if the function is longer and more complex, having a pre-initialized variable to default (say FALSE) and then overriding it on discovery of the param being set is the right thing to do

webrtc-c/canary/src/Include.h
lherman-cs: That was somewhat my plan. My vision is that we'll have a script that will enumerate the permutation that we want to test. By having this script, it'll also expose the possible environment variables to the customers.

webrtc-c/canary/CMakeLists.txt
lherman-cs: Yep, makes sense

webrtc-c/canary/src/Main.cpp
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+CleanUp:
+    deinitKvsWebRtc();
+    Canary::Cloudwatch::deinit();
+
+    return retStatus;
+}
+
+STATUS onNewConnection(std::shared_ptr<Canary::Peer::Connection> pConnection)
```
MushMal, line 68: Need to think through smart pointers usage and their proliferation.

webrtc-c/canary/CMakeLists.txt
lherman-cs: We can definitely add this when we actually need the API. But, so far, I haven't found the need for it.

webrtc-c/canary/src/Main.cpp
```diff
@@ -0,0 +1,151 @@
...
+        // Also, it's very unlikely to have a delay greater than SAMPLE_VIDEO_FRAME_DURATION, so the logic assumes that this is always
+        // true for simplicity.
+        elapsed = lastFrameTime - startTime;
+        THREAD_SLEEP(frameDuration - elapsed % frameDuration);
+        lastFrameTime = GETTIME();
+    }
+
+CleanUp:
+
+    MEMFREE(frame.frameData);
```
MushMal, line 143: Perhaps this should be SAFE_MEMFREE

lherman-cs:
> My only concern about a complete C++ solution is if customers would like to run canaries of their own, they will have to modify their files to a great degree to adapt to a C++ solution, which might be a lot of changes on their end.

Not sure if it's an issue. Since we're using Cloudwatch, we need to use C++ as there's no AWS C sdk. So, the customers still need to use C++ compiler. If they want to adapt our Canary, I think that they can simply clone our repo and add some C since C++ should be somewhat compatible with C.

webrtc-c/canary/src/Cloudwatch.cpp
```diff
@@ -0,0 +1,73 @@
...
+VOID Cloudwatch::deinit()
+{
+    auto& instance = getInstance();
+    instance.logs.deinit();
+    instance.monitoring.deinit();
+}
+
+VOID Cloudwatch::logger(UINT32 level, PCHAR tag, PCHAR fmt, ...)
+{
+    CHAR logFmtString[MAX_LOG_FORMAT_LENGTH + 1];
```
MushMal, line 53: The logger functionality is overridden to this function by setting globalCustomLogPrintFn function pointer. This means that:

1) We won't get any local logs printed into standard output which might be very useful
2) VERY IMPORTANT: There are cases where after failing to flush for example, it still calls DLOG which will result in this function call, which, in turn, will call flush again resulting in stack overflow. This will be impossible to debug.

Recommend:

Adding logic to print out the info into the standard output as well as the cw logs. Also, whenever the flushing or that type of critical path fails, we ensure that we don't use the logger macros to output the failed logging.

Perhaps, in that case, we should error out of the canary application all together.

webrtc-c/canary/src/Peer.cpp
```diff
@@ -0,0 +1,191 @@
...
+    channelInfo.pCertPath = (PCHAR) pConfig->pCertPath;
+    channelInfo.messageTtl = 0; // Default is 60 seconds
+
+    clientCallbacks.customData = (UINT64) this;
+    clientCallbacks.errorReportFn = [](UINT64 customData, STATUS status, PCHAR msg, UINT32 msgLen) -> STATUS {
+        PPeer pPeer = (PPeer) customData;
+        DLOGW("Signaling client generated an error 0x%08x - '%.*s'", status, msgLen, msg);
+
+        UNUSED_PARAM(pPeer);
+        // TODO: handle recreate signaling
```
MushMal, line 93: This is a big one

webrtc-c/canary/src/Cloudwatch.cpp
```diff
@@ -0,0 +1,73 @@
+#include "Include.h"
+
+namespace Canary {
+
+Cloudwatch::Cloudwatch(Canary::PConfig pConfig, ClientConfiguration* pClientConfig) : logs(pConfig, pClientConfig), monitoring(pConfig, pClientConfig)
+{
+}
+
+STATUS Cloudwatch::init(Canary::PConfig pConfig)
+{
```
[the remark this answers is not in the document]
lherman-cs: Makes sense

webrtc-c/canary/src/CloudwatchLogs.cpp
lherman-cs: Yep

webrtc-c/canary/src/Cloudwatch.h
lherman-cs: When I created it, it was meant for logs and monitoring. But, it doesn't apply to monitoring anymore. 

Yes, this is a good idea.

webrtc-c/canary/src/Config.cpp
lherman-cs: So that we don't need to set all the values. But, you're right, this is for canaries, so it's much better to be more strict about this.

webrtc-c/canary/src/Config.cpp
lherman-cs: Makes sense. But, since I've modified this to use mustenv, I have to return STATUS instead. Now, I have it as normal pattern with `CHK_*` and `CleanUp`

webrtc-c/canary/src/Main.cpp
lherman-cs: Yep

webrtc-c/canary/src/Cloudwatch.cpp
lherman-cs: 1. It will still print out to the stdout, it's done through `vprintf(logFmtString, valist);` before pushing the logs to Cloudwatch.
2. It shouldn't cause stackoverflow, the current behavior is that we'll lose some logs with the size of `MAX_CLOUDWATCH_LOG_COUNT` (128). This is the case because every flush call, it'll make a temporary copy of the current buffer (for avoiding a race condition) and clear the current buffer. The next DLOG will only append the log to an empty buffer, it'll call flush again after the buffer is full. 

I totally agree to not use DLOG when the critical path fails. 

I think we should keep the application running, since the logs themselves are still preserved in Jenkins, and pushing logs to Cloudwatch is not a critical path for our Canary. Even if we restarted the canary all together, I don't believe that we won't get into the error state again.

webrtc-c/canary/src/Cloudwatch.h
lherman-cs: Yep, makes sense

webrtc-c/canary/src/Cloudwatch.cpp
```diff
@@ -0,0 +1,81 @@
...
+    ClientConfiguration clientConfig;
+    CreateLogGroupRequest createLogGroupRequest;
+    Aws::CloudWatchLogs::Model::CreateLogStreamOutcome createLogStreamOutcome;
+    CreateLogStreamRequest createLogStreamRequest;
+
+    clientConfig.region = pConfig->pRegion;
+    auto& instance = getInstanceImpl(pConfig, &clientConfig);
+
+    if (STATUS_FAILED(instance.logs.init())) {
+        DLOGW("Failed to create Cloudwatch logger, fallback to local output");
```
disa6302, line 23: I think our fallback should be file logger like we do producer-c. This is because, we do want to preserve logs in files to pull out. The apps we mostly be run as background on CI/CD and we might not have a backup.

webrtc-c/canary/src/Cloudwatch.cpp
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+        globalCustomLogPrintFn = logger;
```
disa6302, line 25: Can you rename this cloudwatchLogger()? Would like it to be consistent with the producer C canary

webrtc-c/canary/src/Cloudwatch.cpp
lherman-cs: @disa6302 since we run this on Jenkins, we can simply get the log by fetching `${BUILD_URL}/consoleText`. I'm trying to keep our Canary as simple as possible, so that we can make it really robust for our alarms, I don't want to include things that can be done from outside.

Even if the user doesn't use Jenkins, they can easily just use `tee` to store the logs to a file.

webrtc-c/canary/src/Cloudwatch.cpp
lherman-cs: It sounds awkward to me, since `logger` here is already namespaced by `Cloudwatch`. If we want to get the global symbol (assuming we're using Canary namespace), it would be `Cloudwatch::logger`. If we renamed this to `cloudwatchLogger`, it becomes `Cloudwatch::cloudwatchLogger`. I think It's similar to Producer C's naming, the only difference is that the namespace here is from C++ instead of a manual namespace.

webrtc-c/canary/src/Main.cpp
lherman-cs: I removed all smart pointers in favor of plain C pointer (except the callbacks from Cloudwatch). This is to make the code simpler, since our SDK is plain C, it's easier to integrate.

webrtc-c/canary/src/Cloudwatch.cpp
disa6302: Ok. Fair.

webrtc-c/canary/src/Cloudwatch.cpp
disa6302: Doesnt `tee` just store what is on the console? What about when the logs run for days? 
Also, if we run the application as a background process, the logs would not appear on the Jenkins console as well. Would ${BUILD_URL}/consoleTest store logs that do not show up on the Jenkins console?

IMO, adding a fileLogger option is not going to add any complexity. We can skip it for now, but, I think we should not rely on external mechanisms for logging.

lherman-cs: In that case, we can just use `split` (it allows you to break up big input to chunks, you can even set the size limit of each chunk):

Screen Shot 2020-08-13 at 1 05 08 PM

Why would we run the application as a background process? Why can't we just spawn a new job and run the application in that job?

lherman-cs: Discussed offline with @disa6302 , we made a decision to add the file logger.

webrtc-c/canary/CMakeLists.txt
lherman-cs: done

webrtc-c/canary/src/Peer.cpp
lherman-cs: Instead of recreating the signaling, I went with the "let it crash" philosophy, and capture the error code as a part of metrics.

webrtc-c/canary/src/Cloudwatch.cpp
lherman-cs: Added file logger.

webrtc-c/canary/src/Main.cpp
```diff
@@ -0,0 +1,179 @@
...
+        elapsed = lastFrameTime - startTime;
+        THREAD_SLEEP(frameDuration - elapsed % frameDuration);
+        lastFrameTime = GETTIME();
+    }
+
+CleanUp:
+
+    SAFE_MEMFREE(frame.frameData);
+
+    auto threadKind = kind == MEDIA_STREAM_TRACK_KIND_VIDEO ? "video" : "audio";
```
disa6302, line 173: Curious why we cannot use kind directly.

lherman-cs: It's just for the sake of readability. By doing this, I can describe my intention through the variable name. Also, I don't need to duplicate the logic for the else statement.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -0,0 +1,73 @@
...
+    this->channelDimension.SetValue(pConfig->pChannelName);
+
+    return retStatus;
+}
+
+VOID CloudwatchMonitoring::deinit()
+{
+    // need to wait all metrics to be flushed out, otherwise we'll get a segfault.
+    // https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/basic-use.html
+    // TODO: maybe add a timeout? But, this might cause a segfault if it hits a timeout.
```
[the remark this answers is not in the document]
lherman-cs: That makes sense. We can definitely use that to properly flush with a timeout.

PR #28, opened 2020-08-14
Add signaling and ICE init delay metrics

lherman-cs:
* Signaling init delay measures the time it takes from Signaling creation to connected
* ICE hole punching delay measures The delay from ICE_AGENT_STATE_CHECK_CONNECTION to ICE_AGENT_STATE_READY

### DIFF
--- a/webrtc-c/canary/src/Peer.cpp
+++ b/webrtc-c/canary/src/Peer.cpp
@@ -241,6 +241,15 @@ STATUS Peer::initPeerConnection()
         DLOGI("New connection state %u", newState);
 
         switch (newState) {
+            case RTC_PEER_CONNECTION_STATE_CONNECTING:
+                pPeer->iceHolePunchingStartTime = GETTIME();
+                break;
+            case RTC_PEER_CONNECTION_STATE_CONNECTED: {
+                auto duration = (GETTIME() - pPeer->iceHolePunchingStartTime) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+                DLOGI("ICE hole punching took %lu ms", duration);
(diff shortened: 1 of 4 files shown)


webrtc-c/canary/src/Peer.cpp
```diff
@@ -61,6 +61,31 @@ STATUS Peer::initSignaling()
...
+    clientCallbacks.stateChangeFn = [](UINT64 customData, SIGNALING_CLIENT_STATE state) -> STATUS {
+        STATUS retStatus = STATUS_SUCCESS;
+        PPeer pPeer = (PPeer) customData;
+        PCHAR pStateStr;
+
+        signalingClientGetStateString(state, &pStateStr);
+        DLOGD("Signaling client state changed to %d - '%s'", state, pStateStr);
+
+        switch (state) {
+            case SIGNALING_CLIENT_STATE_NEW:
```
MushMal, line 73: We need to be more granular perhaps here. What you are measuring is the time it took from the creation of the client to the connection. We have two APIs - create and connect. True, most applications use those in sequence so perhaps we should be OK to mark those as one. We might break it down later into create (with all the state transitions) and connect itself. But, for now, I think we are OK

MushMal: Also, how are we handling the case when it doesn't connect and times out?

lherman-cs: In my head, I would let it to not post any metrics, since timeout means no connection. If we post some metrics, the value is going to be misleading. 

By not posting metrics, we can capture this as no metric data, which will disambiguate between healthy and unhealthy cases. But, I would not set an alarm against this since the "ExitStatus" metrics should also capture this case as well.

PR #31, opened 2020-08-26
Add timeout duration

lherman-cs:
Add a an environment variable to set a timeout duration. `0` means it never timeouts.


webrtc-c/canary/src/Config.cpp
```diff
@@ -96,6 +112,9 @@ STATUS Config::init(INT32 argc, PCHAR argv[], Canary::PConfig pConfig)
                  pConfig->isMaster ? "master" : "viewer", GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
     }
 
+    CHK_STATUS(mustenvUint64(CANARY_DURATION_IN_SECONDS_ENV_VAR, &durationInSeconds));
```
MushMal, line 115: So you are making this variable mandatory? Shouldn't it assume some default like 0 if not specified?

lherman-cs: Yes, I think it's better to error out than suddenly find that the canary doesn't terminate. Since this is one of the important variable, I'd like to make the definition to be explicit from the caller. In my opinion, making this explicit will also help with readability on the caller side.

PR #34, opened 2020-09-04
Rtp stats and end to end metrics

disa6302:
*Issue #, if available:*

*Description of changes:*

- Added RTP Stats (Incoming and outgoing)
- Added end to end metrics (end to end frame latency and data match). These metrics are more of a placeholder now because it does not work as expected. 

Todo:
- Frame size does not match from sender to receiver (there is a difference of 4 bytes end to end - need to investigate)
- End to end latency gives really large values at times because the data in pFrame metadata section is corrupted. Will investigate it

### DIFF
--- a/webrtc-c/canary/src/CanaryWebrtc.cpp
+++ b/webrtc-c/canary/src/CanaryWebrtc.cpp
@@ -4,6 +4,9 @@ STATUS onNewConnection(Canary::PPeer);
 STATUS run(Canary::PConfig);
 VOID sendLocalFrames(Canary::PPeer, MEDIA_STREAM_TRACK_KIND, const std::string&, UINT64, UINT32);
 VOID sendCustomFrames(Canary::PPeer, MEDIA_STREAM_TRACK_KIND, UINT64, UINT64);
+STATUS canaryRtpOutboundStats(UINT32, UINT64, UINT64);
+STATUS canaryRtpInboundStats(UINT32, UINT64, UINT64);
+STATUS canaryEndToEndStats(UINT32, UINT64, UINT64);
 
 std::atomic<bool> terminated;
 VOID handleSignal(INT32 signal)
@@ -11,6 +14,7 @@ VOID handleSignal(INT32 signal)
     UNUSED_PARAM(signal);
     terminated = TRUE;
 }
+
 
 // add frame pts, original frame size, CRC to beginning of buffer after Annex-B format NALu
 VOID addCanaryMetadataToFrameData(PBYTE buffer, PFrame pFrame)
@@ -23,7 +27,7 @@ VOID addCanaryMetadataToFrameData(PBYTE buffer, PFrame pFrame)
     putUnalignedInt32BigEndian((PINT32) pCurPtr, COMPUTE_CRC32(buffer, pFrame->size));
 }
 
-// Frame Data format: NALu (4 bytes) + Header (PTS, Size (including header), CRC (including header) + Randomly generated frameBits
+// Frame Data format: NALu (4 bytes) + Header (PTS, Size (including header), CRC (frame data) + Randomly generated frameBits
 
 // TODO: Support dynamic random frame sizes to bring it closer to real time scenarios
 VOID createCanaryFrameData(PBYTE buffer, PFrame pFrame)
@@ -31,10 +35,10 @@ VOID createCanaryFrameData(PBYTE buffer, PFrame pFrame)
     UINT32 i;
     // For decoding purposes, the first 4 bytes need to be a NALu
     putUnalignedInt32BigEndian((PINT32) buffer, 0x00000001);
-    addCanaryMetadataToFrameData(buffer, pFrame);
     for (i = ANNEX_B_NALU_SIZE + CANARY_METADATA_SIZE; i < pFrame->size; i++) {
         buffer[i] = RAND();
     }
+    addCanaryMetadataToFrameData(buffer, pFrame);
 }
 
 INT32 main(INT32 argc, CHAR* argv[])
@@ -120,7 +124,14 @@ STATUS run(Canary::PConfig pConfig)
 
         // Since the goal of the canary is to test robustness of the SDK, there is not an immediate need
         // to send audio frames as well. It can always be added in if needed in the future
-        std::thread videoThread(sendCustomFrames, &peer, MEDIA_STREAM_TRACK_KIND_VIDEO, pConfig->bytesPerSecond, pConfig->frameRate);
+        std::thread videoThread(sendCustomFrames, &peer, MEDIA_STREAM_TRACK_KIND_VIDEO, pConfig->bitRate, pConfig->frameRate);
+        // All metrics tracking will happen on a time queue to simplify handling periodicity
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpOutboundStats, (UINT64) &peer,
+                                      &timeoutTimerId));
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpInboundStats, (UINT64) &peer,
+                                      &timeoutTimerId));
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, END_TO_END_METRICS_INVOCATION_PERIOD, END_TO_END_METRICS_INVOCATION_PERIOD,
+                                      canaryEndToEndStats, (UINT64) &peer, &timeoutTimerId));
         videoThread.join();
         CHK_STATUS(peer.shutdown());
     }
@@ -204,7 +215,7 @@ VOID sendCustomFrames(Canary::PPeer pPeer, MEDIA_STREAM_TRACK_KIND kind, UINT64 
         // This re-alloc is done in case the estimated size does not match the actual requirement.
         // We do not want to constantly malloc within a loop. Hence, we re-allocate only if required
         // Either ways, the realloc should not happen
-        if (hexStrLen != SIZEOF(frame.frameData)) {
+        if (hexStrLen != (frameSizeWithoutNalu * 2 + 1)) {
             DLOGW("Re allocating...this should not happen...something might be wrong");
             frame.frameData = (PBYTE) REALLOC(frame.frameData, hexStrLen + ANNEX_B_NALU_SIZE);
             CHK_ERR(frame.frameData != NULL, STATUS_NOT_ENOUGH_MEMORY, "Failed to realloc media buffer");
@@ -229,6 +240,45 @@ CleanUp:
     } else {
         DLOGI("%s thread exited successfully", threadKind);
     }
+}
+
+STATUS canaryRtpOutboundStats(UINT32 timerId, UINT64 currentTime, UINT64 customData)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    if (!terminated.load()) {
+        Canary::PPeer pPeer = (Canary::PPeer) customData;
+        pPeer->publishStatsForCanary(RTC_STATS_TYPE_OUTBOUND_RTP);
+    } else {
+        retStatus = STATUS_TIMER_QUEUE_STOP_SCHEDULING;
+    }
+CleanUp:
+    return retStatus;
+}
+
+STATUS canaryRtpInboundStats(UINT32 timerId, UINT64 currentTime, UINT64 customData)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    if (!terminated.load()) {
+        Canary::PPeer pPeer = (Canary::PPeer) customData;
+        pPeer->publishStatsForCanary(RTC_STATS_TYPE_INBOUND_RTP);
+    } else {
+        retStatus = STATUS_TIMER_QUEUE_STOP_SCHEDULING;
+    }
+CleanUp:
+    return retStatus;
+}
+
+STATUS canaryEndToEndStats(UINT32 timerId, UINT64 currentTime, UINT64 customData)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    if (!terminated.load()) {
+        Canary::PPeer pPeer = (Canary::PPeer) customData;
+        pPeer->publishEndToEndMetrics();
+    } else {
+        retStatus = STATUS_TIMER_QUEUE_STOP_SCHEDULING;
+    }
+CleanUp:
+    return retStatus;
 }
 
 VOID sendLocalFrames(Canary::PPeer pPeer, MEDIA_STREAM_TRACK_KIND kind, const std::string& pattern, UINT64 frameCount, UINT32 frameDuration)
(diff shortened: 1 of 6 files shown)


webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
     this->push(datum);
 }
 
+VOID CloudwatchMonitoring::pushOutboundRtpStats(Canary::POutgoingRTPMetricsContext pOutboundRtpStats)
+{
+    MetricDatum bytesDiscardedPercentageDatum, averageFramesRateDatum, nackRateDatum, retransmissionPercentDatum;
+
+    bytesDiscardedPercentageDatum.SetMetricName("Percentageframediscarded");
```
lherman-cs, line 102: Percentageframediscarded -> PercentageFrameDiscarded

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
...
+    bytesDiscardedPercentageDatum.AddDimensions(this->channelDimension);
+    this->push(bytesDiscardedPercentageDatum);
+
+    averageFramesRateDatum.SetMetricName("FramesPerSecond");
+    averageFramesRateDatum.SetValue(pOutboundRtpStats->averageFramesSentPerSecond);
+    averageFramesRateDatum.SetUnit(StandardUnit::Count_Second);
+    averageFramesRateDatum.AddDimensions(this->channelDimension);
+    this->push(averageFramesRateDatum);
+
+    nackRateDatum.SetMetricName("NackRate");
```
lherman-cs, line 114: Probably NackPerSecond instead to be more consistent with the rest?

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
...
+    this->push(retransmissionPercentDatum);
+}
+
+VOID CloudwatchMonitoring::pushInboundRtpStats(Canary::PIncomingRTPMetricsContext pIncomingRtpStats)
+{
+    MetricDatum incomingBitrateDatum, incomingPacketRate, incomingFrameDropRateDatum;
+
+    incomingBitrateDatum.SetMetricName("IncomingBitRate");
+    incomingBitrateDatum.SetValue(pIncomingRtpStats->incomingBitRate);
+    incomingBitrateDatum.SetUnit(StandardUnit::Kilobits_Second);
```
lherman-cs, line 133: Is this supposed to be Bits_second instead?

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
...
+    incomingBitrateDatum.AddDimensions(this->channelDimension);
+    this->push(incomingBitrateDatum);
+
+    incomingPacketRate.SetMetricName("IncomingPacketRate");
+    incomingPacketRate.SetValue(pIncomingRtpStats->packetReceiveRate);
+    incomingPacketRate.SetUnit(StandardUnit::Count_Second);
+    incomingPacketRate.AddDimensions(this->channelDimension);
+    this->push(incomingPacketRate);
+
+    incomingFrameDropRateDatum.SetMetricName("IncomingFrameDropRate");
```
lherman-cs, line 143: How about IncomingFramesDroppedPerSecond? same reason as above.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
...
+{
+    MetricDatum incomingBitrateDatum, incomingPacketRate, incomingFrameDropRateDatum;
+
+    incomingBitrateDatum.SetMetricName("IncomingBitRate");
+    incomingBitrateDatum.SetValue(pIncomingRtpStats->incomingBitRate);
+    incomingBitrateDatum.SetUnit(StandardUnit::Kilobits_Second);
+    incomingBitrateDatum.AddDimensions(this->channelDimension);
+    this->push(incomingBitrateDatum);
+
+    incomingPacketRate.SetMetricName("IncomingPacketRate");
```
lherman-cs, line 137: How about IncomingPacketsPerSecond? same reason as above.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -96,4 +95,82 @@ VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, StandardUnit u
...
+    // This debug will be removed in the future
+    for (auto it = pEndToEndMetricsContext->frameLatency.begin(); it != pEndToEndMetricsContext->frameLatency.end(); it++) {
+        DLOGD("Latency: %lf", *it);
+    }
+    DLOGD("Vector size: %lu", pEndToEndMetricsContext->frameLatency.size());
+    endToEndLatencyDatum.SetValues(pEndToEndMetricsContext->frameLatency);
+    this->push(endToEndLatencyDatum);
+
+    dataMatchDatum.SetMetricName("FrameDataMatch");
+    dataMatchDatum.SetStorageResolution(1);
```
lherman-cs, line 166: Why do we need to change the resolution to 1 second?

webrtc-c/canary/src/Main.cpp
```diff
@@ -86,11 +100,20 @@ STATUS run(Canary::PConfig pConfig)
...
+        std::thread audioThread(sendLocalFrames, &peer, MEDIA_STREAM_TRACK_KIND_AUDIO, CANARY_AUDIO_FRAMES_PATH, NUMBER_OF_OPUS_FRAME_FILES,
+                                SAMPLE_AUDIO_FRAME_DURATION);
+
+        // All metrics tracking will happen on a time queue to simplify handling periodicity
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpOutboundStats, (UINT64) &peer,
+                                      &timeoutTimerId));
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpInboundStats, (UINT64) &peer,
+                                      &timeoutTimerId));
+        if (!pConfig->isMaster) {
+            CHK_STATUS(timerQueueAddTimer(timerQueueHandle, END_TO_END_METRICS_INVOCATION_PERIOD, END_TO_END_METRICS_INVOCATION_PERIOD,
```
lherman-cs, line 114: Since each peer emits video and audio, I think it makes sense for both peers to emit the same set of metrics.

webrtc-c/canary/src/Main.cpp
```diff
@@ -86,11 +100,20 @@ STATUS run(Canary::PConfig pConfig)
...
+                                          canaryEndToEndStats, (UINT64) &peer, &timeoutTimerId));
+        }
         videoThread.join();
         audioThread.join();
         CHK_STATUS(peer.shutdown());
```
lherman-cs, line 119: Important: the metric timers need to be stopped before shutting down peer. Otherwise, it'll try to use peer that has been shutdown and might cause a crash.

Probably guard each callback by `terminated.load()`. If it's terminated, you can return the timer cancellation status.

disa6302: Good point. Done.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
disa6302: I had added it earlier because I was testing emitting metrics at a 1 second interval and I wanted to set high resolution. Don't need it anymore. Removed.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
disa6302: Fixed all naming.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
disa6302: It is supposed to be Kilobits_second. I corrected the related computation

webrtc-c/canary/src/Peer.cpp
```diff
@@ -533,13 +555,145 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
...
+    DOUBLE videoFramesGenerated = 0;
+    DOUBLE videoBytesGenerated = 0;
+
+    {
+        std::lock_guard<std::mutex> lock(this->countUpdateMutex);
+        videoFramesGenerated = (DOUBLE) this->canaryOutgoingRTPMetricsContext.videoFramesGenerated;
+        videoBytesGenerated = (DOUBLE) this->canaryOutgoingRTPMetricsContext.videoBytesGenerated;
     }
 
+    currentDuration = (DOUBLE)(this->canaryMetrics.timestamp - this->canaryOutgoingRTPMetricsContext.prevTs) / HUNDREDS_OF_NANOS_IN_A_SECOND;
```
lherman-cs, line 603: Important: since this block is not guarded with a `countUpdateMutex` lock, it's possible that we'll lose some data for `videoFramesGenerated` and `videoBytesGenerated`. Between 603 and 616, it's possible that there's a new call to `writeFrame` and increment these variables.

webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -91,11 +105,18 @@ STATUS run(Canary::PConfig pConfig)
...
-        std::thread audioThread(sendLocalFrames, &peer, MEDIA_STREAM_TRACK_KIND_AUDIO, "./assets/opusSampleFrames/sample-%03d.opus",
-                                NUMBER_OF_OPUS_FRAME_FILES, SAMPLE_AUDIO_FRAME_DURATION);
-
+        std::thread videoThread(sendLocalFrames, &peer, MEDIA_STREAM_TRACK_KIND_VIDEO, CANARY_VIDEO_FRAMES_PATH, NUMBER_OF_H264_FRAME_FILES,
+                                SAMPLE_VIDEO_FRAME_DURATION);
+        std::thread audioThread(sendLocalFrames, &peer, MEDIA_STREAM_TRACK_KIND_AUDIO, CANARY_AUDIO_FRAMES_PATH, NUMBER_OF_OPUS_FRAME_FILES,
+                                SAMPLE_AUDIO_FRAME_DURATION);
+
+        // All metrics tracking will happen on a time queue to simplify handling periodicity
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpOutboundStats, (UINT64) &peer,
```
lherman-cs, line 114: For simplicity, I feel like we should combine all of these 3 callbacks into a single callback. `END_TO_END_METRICS_INVOCATION_PERIOD` can be equal to `METRICS_INVOCATION_PERIOD`.

Is there any reason why they're separated?

webrtc-c/canary/src/Peer.cpp
```diff
@@ -533,13 +555,145 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
...
+    DLOGD("Packet receive rate: %lf", this->canaryIncomingRTPMetricsContext.packetReceiveRate);
+    DLOGD("Incoming bit rate: %lf", this->canaryIncomingRTPMetricsContext.incomingBitRate / 1024.0);
+    DLOGD("Frame drop rate: %lf", this->canaryIncomingRTPMetricsContext.framesDroppedPerSecond);
+CleanUp:
+    return retStatus;
+}
+STATUS Peer::publishStatsForCanary()
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    std::lock_guard<std::recursive_mutex> lock(this->mutex);
```
lherman-cs, line 671: Why do we need to acquire a lock?

webrtc-c/canary/src/Peer.cpp
```diff
@@ -533,13 +555,145 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
...
+            CHK(FALSE, STATUS_NOT_IMPLEMENTED);
+    }
 CleanUp:
+    return retStatus;
+}
 
+STATUS Peer::publishEndToEndMetrics()
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    std::lock_guard<std::recursive_mutex> lock(this->mutex);
```
lherman-cs, line 694: Same here

webrtc-c/canary/src/Peer.cpp
```diff
@@ -487,29 +496,42 @@ STATUS Peer::handleSignalingMsg(PReceivedSignalingMessage pMsg)
...
+        // latency metrics. Need to root cause the issue
+        PPeer pPeer = (Canary::PPeer)(customData);
+        std::unique_lock<std::recursive_mutex> lock(pPeer->mutex);
+        PBYTE pCurPtr = pFrame->frameData + pFrame->size - (UINT32) CANARY_METADATA_SIZE;
+        UINT32 size = getUnalignedInt32BigEndian((PINT32)(pCurPtr));
+        pCurPtr += SIZEOF(UINT32);
+        UINT64 tsRx = getUnalignedInt64BigEndian((PINT64)(pCurPtr));
+        pCurPtr += SIZEOF(UINT64);
+        UINT32 crc = getUnalignedInt32BigEndian((PINT32)(pCurPtr));
+        pPeer->endToEndMetricsContext.dataMatch.push_back(crc == COMPUTE_CRC32(pFrame->frameData, pFrame->size) ? 1.0 : 0.0);
```
lherman-cs, line 516: I think EndToEndMetricsContext should have its own lock, so that we Peer's lock is acquired the media thread doesn't block. I'm fine either way, but I feel like mutexes should be used for the smallest scope possible so that we can get the most benefit from parallelism.

disa6302: I think this should be ok. Because the callback shouldnt take long to execute.

webrtc-c/canary/src/Peer.cpp
disa6302: Actually, now that I think about I dont need it. Removing.

webrtc-c/canary/src/CanaryWebrtc.cpp
disa6302: The reason is if at any point of time, I want to change the period I extract each metrics, it would be easier if these were separate tasks. It is unlikely at this point, but just wanted to cover the case. We can always compress this if we need to later.

webrtc-c/canary/src/Peer.cpp
disa6302: Agreed. Removed `videoFramesGenerated` and `videoBytesGenerated` and I am guarding the lines with the mutex.

webrtc-c/canary/src/Peer.cpp
lherman-cs: Fair point

webrtc-c/canary/src/CanaryWebrtc.cpp
lherman-cs: I see That's fair

lherman-cs (approved):
LGTM

lherman-cs (approved):

PR #44, opened 2020-09-16
Introduce custom frame format

disa6302:
*Issue #, if available:*

*Description of changes:*

- Included frame rate and data rate configurability. `Frame size` is calculated as `data_rate(in bytes) / frame_rate`, hence, both are mandatory arguments. 
- Added custom frame format, replacing the use of H264 files. Currently, only video frames are added since there is not going to be any additional data point/metrics with audio frames. 

Todo:

Support dynamic frame size changes to mimic real world scenarios


webrtc-c/canary/src/Include.h
```diff
@@ -31,6 +31,11 @@
 #define CANARY_LOG_STREAM_NAME_ENV_VAR     "CANARY_LOG_STREAM_NAME"
 #define CANARY_CERT_PATH_ENV_VAR           "CANARY_CERT_PATH"
 #define CANARY_DURATION_IN_SECONDS_ENV_VAR "CANARY_DURATION_IN_SECONDS"
+#define CANARY_BYTE_RATE_ENV_VAR           "CANARY_DATARATE_IN_BYTES_PERSECOND"
+#define CANARY_FRAME_RATE_ENV_VAR          "CANARY_FRAME_RATE"
+
+#define CANARY_METADATA_SIZE               (SIZEOF(UINT64) + SIZEOF(UINT32) + SIZEOF(UINT32))
```
MushMal, line 37: Instead of this, we can define a structure in C and simply use SIZEOF(CanaryMetadata);

webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -149,48 +172,20 @@ STATUS onNewConnection(Canary::PPeer pPeer)
...
-    CHAR filePath[MAX_PATH_LEN + 1];
-    UINT64 startTime, lastFrameTime, elapsed;
-
-    frame.frameData = NULL;
-    frame.size = 0;
-    frame.presentationTs = 0;
-    startTime = GETTIME();
-    lastFrameTime = startTime;
-
+    frame.size = CANARY_METADATA_SIZE + (dataRate / frameRate);
```
MushMal, line 179: Recommend having a new structure that wraps/encapsulates Frame like so

typedef struct {
    UINT64 field1;
    UINT64 field2;
...
} CanaryFrameMetadata;

typedef struct {
    Frame frame;
    CanaryFrameMetadata canaryMetadata;
    // -> actual bits of the frame following the matadata;
} CanaryFrame, *PCanaryFrame;

The allocation would happen something like this

PCanaryFrame pCanaryFrame = (PCanaryFrame) MEMALLOC(SIZEOF(CanaryFrame) + frameSize);
CHK(pCanaryFrame != NULL, STATUS_NOT_ENOUGH_MEMORY);

// Set the fields up
pCanaryFrame->frame.frameData = (PBYTE)(pCanaryFrame + 1);
pCanaryFrame->frame.frameSize = frameSize + SIZEOF(CanaryFrameMetadata);
...

MushMal: This will work directly if both sides are on the same machine with the same packing and endianness. 
Care must be taken to handle packing and endianness.

For now, using direct packing into the buffer is a better option I think

webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -11,15 +11,41 @@ VOID handleSignal(INT32 signal)
     terminated = TRUE;
 }
 
+// add frame pts, original frame size, CRC to beginning of buffer after Annex-B format NALu
```
MushMal, line 14: Care must be taken to avoid zeroes. Recommend encoding the metadata into base64 to avoid NALu escaping

MushMal (commented):
I added a lot of comment about using C structures instead of packing the buffer. I think you should just ignore that as it's actually better to deal with the bits.

My only real comment is about the NALu and zeroes. I think the best would be to have a metadata struct that handles serialization and deserialization into a string. The underlying format then will be base64 encoded (or hex encoded - even better)

webrtc-c/canary/src/Include.h
disa6302: As per your comment, and I think the same, it is better to deal with the actual bits rather than with a structure at this point, owing to endianness across platforms.

webrtc-c/canary/src/CanaryWebrtc.cpp
disa6302: Makes sense! Fixed.

webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -149,48 +172,39 @@ STATUS onNewConnection(Canary::PPeer pPeer)
     return retStatus;
 }
 
-VOID sendLocalFrames(Canary::PPeer pPeer, MEDIA_STREAM_TRACK_KIND kind, const std::string& pattern, UINT64 frameCount, UINT32 frameDuration)
+VOID sendCustomFrames(Canary::PPeer pPeer, MEDIA_STREAM_TRACK_KIND kind, UINT64 dataRate, UINT64 frameRate)
```
MushMal, line 175: I think we should preserve both of the functions. We might still want to get the pseudo-video frames with some combination of params to the canary

MushMal (commented):
Mostly about not removing the frame sending function and doing more streamlined in-place hex encoding

MushMal:
BTW, where do we hexDecode the frame data to compare stuff

webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -149,48 +172,39 @@ STATUS onNewConnection(Canary::PPeer pPeer)
...
-
-        frame.presentationTs += frameDuration;
-
+        frame.size = actualFrameSize;
+        createCanaryFrameData(&frame);
+
+        // Hex encode the data (without the ANNEX-B NALu) to ensure parts of random frame data is not skipped if they
+        // are the same as the ANNEX-B NALu
+        CHK_STATUS(hexEncode(frame.frameData + ANNEX_B_NALU_SIZE, frameSizeWithoutNalu, NULL, &hexStrLen));
+        CHAR hexStr[hexStrLen];
```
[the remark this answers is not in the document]
disa6302: Fixed this.

disa6302:
> BTW, where do we hexDecode the frame data to compare stuff

We would do that is the transceiverOnFrame callbacks. My RTP Stats PR includes that, except, I will have to rebase these changes on the other PR.

MushMal (approved):

PR #53, opened 2020-09-21
Allow JSON file to initialize config

lherman-cs:
This PR allows users to initialize config with a json file. Environment variables are now being used as a fallback.

Resolves https://github.com/aws-samples/amazon-kinesis-video-streams-demos/issues/56


webrtc-c/canary/src/Config.cpp
```diff
@@ -53,15 +101,14 @@ STATUS mustenvBool(CHAR const* pKey, Config::Value<BOOL>* pResult)
...
-            pResult->value = TRUE;
-        } else {
-            pResult->value = FALSE;
-        }
+    CHK(pResult != NULL, STATUS_NULL_ARG);
+    CHK(!pResult->initialized, retStatus);
+    CHK_STATUS(optenv(pKey, &raw, ""));
+
+    if (raw.value[0] != '\0') {
+        pResult->value = strtobool(raw.value);
```
disa6302, line 111: Nitpick: `strtobool ` -> `strToBool`

webrtc-c/canary/src/Config.cpp
```diff
@@ -201,4 +221,114 @@ STATUS Config::init(INT32 argc, PCHAR argv[])
...
+        } else if (compareJsonString((PCHAR) raw, &tokens[i], JSMN_STRING, (PCHAR) CANARY_ITERATION_IN_SECONDS_ENV_VAR)) {
+            jsonUint64(raw, tokens[++i], &iterationDuration);
+            iterationDuration.value *= HUNDREDS_OF_NANOS_IN_A_SECOND;
+        } else if (compareJsonString((PCHAR) raw, &tokens[i], JSMN_STRING, (PCHAR) CANARY_FORCE_TURN_ENV_VAR)) {
+            jsonBool(raw, tokens[++i], &forceTurn);
+        } else if (compareJsonString((PCHAR) raw, &tokens[i], JSMN_STRING, (PCHAR) CANARY_BIT_RATE_ENV_VAR)) {
+            jsonUint64(raw, tokens[++i], &bitRate);
+        } else if (compareJsonString((PCHAR) raw, &tokens[i], JSMN_STRING, (PCHAR) CANARY_FRAME_RATE_ENV_VAR)) {
+            jsonUint64(raw, tokens[++i], &frameRate);
+        } else if (compareJsonString((PCHAR) raw, &tokens[i], JSMN_STRING, (PCHAR) ACCESS_KEY_ENV_VAR)) {
```
disa6302, line 315: Do we want access key, secret key and session token as jsons? 
I think we should not give a json capability to avoid any accidental git pushes by any customer of the json file.

webrtc-c/canary/src/Config.cpp
lherman-cs: This is intentional. It's consistent with other `strto*` APIs.

webrtc-c/canary/src/Config.cpp
lherman-cs: That's a good point. I'll change that.

lherman-cs: Updated.

webrtc-c/canary/src/CanarySignaling.cpp
```diff
@@ -305,7 +305,7 @@ STATUS run(Canary::PConfig pConfig)
     // Generate a random channel name if not specified in the config.
     // In case we generate the random name we will follow-up with deleting
     // it upon exit to prevent the account from ever increasing channel count
-    if (pConfig->channelName.value == NULL || IS_EMPTY_STRING(pConfig->channelName.value)) {
+    if (pConfig->channelName.value[0] == '\0' || IS_EMPTY_STRING(pConfig->channelName.value)) {
```
MushMal, line 308: Actually, my bad - I forgot that this is an array rather than a pointer. We should just remove the first check and only leave the IS_EMPTY_STRING one

webrtc-c/canary/src/Config.cpp
```diff
@@ -2,29 +2,78 @@
...
+
+    CHK(argv != NULL, STATUS_NULL_ARG);
+    MEMSET(this, 0, SIZEOF(Config));
+
+    if (argc == 2) {
+        DLOGI("Reading configuration from %s\n", argv[1]);
+        CHK_STATUS(initWithJSON(argv[1]));
+    }
+
+    CHK_STATUS(initWithEnvVars());
```
MushMal, line 18: Perhaps move initWithEnvVars before the initWithJSON

webrtc-c/canary/src/Config.cpp
```diff
@@ -2,29 +2,78 @@
...
 {
     STATUS retStatus = STATUS_SUCCESS;
+    const CHAR* value;
 
     CHK(pResult != NULL, STATUS_NULL_ARG);
+    CHK(!pResult->initialized, retStatus);
 
-    CHK_ERR((pResult->value = getenv(pKey)) != NULL, STATUS_INVALID_OPERATION, "%s must be set", pKey);
+    CHK_ERR((value = getenv(pKey)) != NULL, STATUS_INVALID_OPERATION, "%s must be set", pKey);
+    SNPRINTF(pResult->value, ARRAY_SIZE(pResult->value), "%s", value);
```
MushMal, line 57: Not sure I understand the reason for SNPRINF? Why not STRNCPY? Much faster

webrtc-c/canary/src/Config.cpp
```diff
@@ -2,29 +2,78 @@
...
 
     CHK(pResult != NULL, STATUS_NULL_ARG);
+    CHK(!pResult->initialized, retStatus);
 
-    if (NULL == (pResult->value = getenv(pKey))) {
-        pResult->value = pDefault;
+    if (NULL == (value = getenv(pKey))) {
+        value = pDefault;
     }
+    SNPRINTF(pResult->value, ARRAY_SIZE(pResult->value), "%s", value);
```
MushMal, line 76: same

webrtc-c/canary/src/Config.cpp
```diff
@@ -149,48 +172,45 @@ STATUS Config::init(INT32 argc, PCHAR argv[])
...
     CHK_STATUS(optenv(CANARY_CHANNEL_NAME_ENV_VAR, &channelName, CANARY_DEFAULT_CHANNEL_NAME));
     CHK_STATUS(optenv(CANARY_CLIENT_ID_ENV_VAR, &clientId, CANARY_DEFAULT_CLIENT_ID));
     CHK_STATUS(optenvBool(CANARY_IS_MASTER_ENV_VAR, &isMaster, TRUE));
 
-    CHK_STATUS(optenv(CANARY_LOG_GROUP_NAME_ENV_VAR, &logGroupName, CANARY_DEFAULT_LOG_GROUP_NAME));
-    STRNCPY(this->logGroupName.value, logGroupName.value, ARRAY_SIZE(this->logGroupName.value) - 1);
-    this->logGroupName.initialized = TRUE;
+    if (!this->logGroupName.initialized) {
+        CHK_STATUS(optenv(CANARY_LOG_GROUP_NAME_ENV_VAR, &logGroupName, CANARY_DEFAULT_LOG_GROUP_NAME));
+        SNPRINTF(this->logGroupName.value, ARRAY_SIZE(this->logGroupName.value), "%s", logGroupName.value);
```
MushMal, line 191: STRNCPY?

webrtc-c/canary/src/Config.cpp
```diff
@@ -149,48 +172,45 @@ STATUS Config::init(INT32 argc, PCHAR argv[])
...
-    pLogStreamName = getenv(CANARY_LOG_STREAM_NAME_ENV_VAR);
-    if (pLogStreamName != NULL) {
-        STRNCPY(this->logStreamName.value, pLogStreamName, ARRAY_SIZE(this->logStreamName.value) - 1);
-    } else {
-        SNPRINTF(this->logStreamName.value, ARRAY_SIZE(this->logStreamName.value) - 1, "%s-%s-%llu", channelName.value,
-                 isMaster.value ? "master" : "viewer", GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    if (!this->logStreamName.initialized) {
+        pLogStreamName = getenv(CANARY_LOG_STREAM_NAME_ENV_VAR);
+        if (pLogStreamName != NULL) {
+            SNPRINTF(this->logStreamName.value, ARRAY_SIZE(this->logStreamName.value), "%s", logStreamName.value);
```
MushMal, line 198: same

webrtc-c/canary/src/Config.cpp
```diff
@@ -201,4 +221,108 @@ STATUS Config::init(INT32 argc, PCHAR argv[])
...
+          this->trickleIce.value ? "True" : "False", this->useTurn.value ? "True" : "False", this->logLevel.value, this->logGroupName.value,
+          this->logStreamName.value, this->duration.value / HUNDREDS_OF_NANOS_IN_A_SECOND,
+          this->iterationDuration.value / HUNDREDS_OF_NANOS_IN_A_SECOND);
+}
+
+VOID jsonString(PBYTE pRaw, jsmntok_t token, Config::Value<Config::String>* pResult)
+{
+    UINT32 tokenLength = (UINT32)(token.end - token.start);
+
+    SNPRINTF(pResult->value, MIN(tokenLength + 1, ARRAY_SIZE(pResult->value)), "%s", (const CHAR*) pRaw + token.start);
```
MushMal, line 249: STRCPY

webrtc-c/canary/src/Config.cpp
```diff
@@ -201,4 +221,108 @@ STATUS Config::init(INT32 argc, PCHAR argv[])
...
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT64 size;
+    jsmn_parser parser;
+    int r;
+    BYTE raw[MAX_CONFIG_JSON_FILE_SIZE];
+    Config::Value<UINT64> logLevel64;
+
+    CHK_STATUS(readFile(filePath, TRUE, NULL, &size));
+    CHK_ERR(size < 1024, STATUS_INVALID_ARG_LEN, "File size too big. Max allowed is 1024 bytes");
```
MushMal, line 281: MAX_CONFIG_JSON_FILE_SIZE instead of 1024. Same with the printing

MushMal (commented):
minor comments

webrtc-c/canary/src/Config.cpp
lherman-cs: I think the idea is to use JSON as the main values, and use env vars as a fallback. So, the order should be correct as it is. Is there any reason why it should be swapped?

webrtc-c/canary/src/CanarySignaling.cpp
lherman-cs: Ouch, sorry, I just realized that..

webrtc-c/canary/src/Config.cpp
MushMal: You can either do the non-overriding version of the configs and call JSON and then ENV initialization or simply do the overrides in which case env vars should be first and then JSON. I am OK either way - both approaches are OK.

webrtc-c/canary/src/Config.cpp
lherman-cs: I agree that STRNCPY is much faster. But, STRNCPY is also more dangerous and we don't need speed in this code path. 

STRNCPY fails to null terminate if the whole buffer is used for string data (i.e. if the source string length is at least as long as the dest buffer)

For example:

```c
  char raw[] = "hello";
  char buff[2] = "aa";
  
  // buff will not be null terminated
  strncpy(buff, raw, 1);
```

webrtc-c/canary/src/Config.cpp
lherman-cs: Ouch, my bad.

webrtc-c/canary/src/Config.cpp
MushMal: Understood. Normally we reserve a single char more space, copy N and then set the last to NULL. You can also check if it "overflown" the buffer looking at the result. Using STRNCPY is really weird IMHO. However, I am OK if you decide to leave SNPRINTF

webrtc-c/canary/src/Config.cpp
lherman-cs: Updated.

webrtc-c/canary/src/Config.cpp
hassanctech: In this case why not use `STRNCAT` which will *always* null terminate?

webrtc-c/canary/src/CanarySignaling.cpp
lherman-cs: Updated.

webrtc-c/canary/src/Config.cpp
lherman-cs: `STRNCAT` might be more error prone as the last parameter is not the length, but it's the remaining space. More details about `STRNCPY` and `STRNCAT` buffer overflow vulnerabilities, https://us-cert.cisa.gov/bsi/articles/knowledge/coding-practices/strncpy-and-strncat.

webrtc-c/canary/src/Config.cpp
lherman-cs: I think I'll leave it is as it is since the code is already written. We can update this in the future when needed.

MushMal (approved):

disa6302 (approved):
LGTM

webrtc-c/canary/src/Config.cpp
hassanctech: Right I just wanted to present an alternative to STRNCPY/SNPRINTF in case you didn't consider it already.  I have no issues with SNPRINTF.

hassanctech (approved):

Issue #46, opened 2020-09-17
Unify master and viewer into single binary


lherman-cs:
As of now, webrtc canary assumes that master and viewer will run in different nodes, and there's no way to have master and viewer to run in a single process. Having both peers to run in a single process is useful for some scenarios, e.g. measuring TURN end-to-end latency. So, we need to have a configuration that can decide either run both peers in 1 process or not.

PR #57, opened 2020-09-21
Enable running both peers in single process

lherman-cs:
Resolves https://github.com/aws-samples/amazon-kinesis-video-streams-demos/issues/46


webrtc-c/canary/src/CanaryWebrtc.cpp
```diff
@@ -150,6 +153,44 @@ STATUS run(Canary::PConfig pConfig)
...
+    pConfig->print();
+    CHK_STATUS(peer.init(pConfig, callbacks));
+    CHK_STATUS(peer.connect());
+
+    {
+        // Since the goal of the canary is to test robustness of the SDK, there is not an immediate need
+        // to send audio frames as well. It can always be added in if needed in the future
+        std::thread videoThread(sendCustomFrames, &peer, MEDIA_STREAM_TRACK_KIND_VIDEO, pConfig->bitRate.value, pConfig->frameRate.value);
+        // All metrics tracking will happen on a time queue to simplify handling periodicity
+        CHK_STATUS(timerQueueAddTimer(timerQueueHandle, METRICS_INVOCATION_PERIOD, METRICS_INVOCATION_PERIOD, canaryRtpOutboundStats, (UINT64) &peer,
```
MushMal, line 178: Just a note: Alternatively, the handler could simply report all 3 metrics when invoked. However, this works too.

MushMal (approved):

webrtc-c/canary/src/CanaryWebrtc.cpp
lherman-cs: Ah that makes sense as well! I'll keep things as they are for now. We can iterate more later when needed.

Issue #47, opened 2020-09-17
Webrtc Canary should delete signaling channels after terminated


lherman-cs:
Since it's very likely that the canary is run multiple times with different channels, it's important to always clean up these channels to make sure that we don't hit the limit.

MushMal:
The signaling canary itself does delete the channel. We need to apply a similar strategy to others

lherman-cs:
This issue has been fixed by https://github.com/aws-samples/amazon-kinesis-video-streams-demos/pull/65. Closing.

PR #65, opened 2020-10-01
Delete signaling channel during shutdown

lherman-cs:
Since Canary can be used for load testing, it's common to use random channel names. So, it would be nice to delete these channels automatically during shutdown.

Related issue: https://github.com/aws-samples/amazon-kinesis-video-streams-demos/issues/47

### DIFF
--- a/webrtc-c/canary/src/Peer.cpp
+++ b/webrtc-c/canary/src/Peer.cpp
@@ -313,6 +313,10 @@ STATUS Peer::shutdown()
         CHK_LOG_ERR(closePeerConnection(this->pPeerConnection));
     }
 
+    if (!this->isMaster && IS_VALID_SIGNALING_CLIENT_HANDLE(this->pSignalingClientHandle)) {
+        CHK_LOG_ERR(signalingClientDeleteSync(this->pSignalingClientHandle));
+    }
+
     return this->status;
 }
 


webrtc-c/canary/src/Peer.cpp
```diff
@@ -313,6 +313,10 @@ STATUS Peer::shutdown()
         CHK_LOG_ERR(closePeerConnection(this->pPeerConnection));
     }
 
+    if (IS_VALID_SIGNALING_CLIENT_HANDLE(this->pSignalingClientHandle)) {
```
MushMal, line 316: Just verifying that in this case a single peer is the entire canary run so it's OK to delete the entire channel with this peer? I am trying to understand the lifecycle and pardon my ignorance here. Normally, the peers have shorter lifecycle than the signaling channel

MushMal (commented):
More of a question

webrtc-c/canary/src/Peer.cpp
lherman-cs: This should be OK since the peer connection will close first and notify the other peer to also terminate, there shouldn't be anymore communication and this kind of communication should be done in peer-to-peer manner. But, to guarantee this even further, I can add a check so that it's done only in the viewer, similar to what we have for signaling canary.

MushMal (approved):

PR #70, opened 2020-10-02
Reduce storage requirements for webrtc canaries

lherman-cs:
Reduce storage requirements by pruning library dependencies and source codes.

### DIFF
--- a/webrtc-c/canary/CMakeLists.txt
+++ b/webrtc-c/canary/CMakeLists.txt
@@ -6,14 +6,16 @@ set(CMAKE_CXX_STANDARD 11)
 include(FetchContent)
 
 # AWS SDK options
-set(CUSTOM_MEMORY_MANAGEMENT OFF)
-set(ENABLE_TESTING OFF)
-set(BUILD_ONLY "monitoring;logs")
+set(CUSTOM_MEMORY_MANAGEMENT OFF CACHE BOOL "Tell AWS SDK to not use custom memory management" FORCE)
+set(ENABLE_TESTING OFF CACHE BOOL "Disable building tests for AWS SDK" FORCE)
+set(BUILD_ONLY "monitoring;logs" CACHE STRING "Tell AWS SDK to only build monitoring and logs" FORCE)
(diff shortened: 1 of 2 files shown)


webrtc-c/canary/CMakeLists.txt
```diff
@@ -6,14 +6,16 @@ set(CMAKE_CXX_STANDARD 11)
...
+set(ENABLE_TESTING OFF CACHE BOOL "Disable building tests for AWS SDK" FORCE)
+set(BUILD_ONLY "monitoring;logs" CACHE STRING "Tell AWS SDK to only build monitoring and logs" FORCE)
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "Enable static build for AWS SDK" FORCE)
+set(BUILD_STATIC_LIBS ON CACHE BOOL "Enable static build for KVS Webrtc SDK" FORCE)
 
 FetchContent_Declare(
   webrtc
   GIT_REPOSITORY https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c
-  GIT_TAG        master
+  GIT_TAG        35fc3ea5f2132613a8439da41fdfc1b5c052c858
```
disa6302, line 18: In this case, wouldn't we have to keep changing the git tag every time a new commit gets merged in the webrtc repo?

lherman-cs: Well, it's only when we want to update it. Right now, our canary's script doesn't lock in git hash. So, I figure we should do this for now.

lherman-cs: I just want to make sure that we retrieve the version that has the static build fix.

disa6302: Got it. So, this will be changed to master ultimately, correct?

MushMal (approved):

disa6302 (approved):

webrtc-c/canary/CMakeLists.txt
lherman-cs: Yea.. as soon as I finish updating the Groovy script to lock in the version hash.

PR #78, opened 2020-10-05
Lock in commit hash across runners

lherman-cs:
By locking in commit hash, we can guarantee version consistency between runners, as it's determined by a single source of truth (orchestrator).


webrtc-c/canary/jobs/orchestrator.groovy
```diff
@@ -109,6 +108,10 @@ pipeline {
                         script {
                             echo "New runner: ${NEXT_AVAILABLE_RUNNER}"
                             Jenkins.instance.getItemByFullName(NEXT_AVAILABLE_RUNNER).setDisabled(false)
+
+                            // Lock in current commit hash to avoid incosistent version accross runners
```
hassanctech, line 112: inconsistent*

hassanctech (approved):
LGTM!

PR #85, opened 2020-10-07
Add webrtc canary label dimension

lherman-cs:
Add an aggregated metrics by using labels.


webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -13,6 +13,9 @@ STATUS CloudwatchMonitoring::init()
     this->channelDimension.SetName("Channel");
     this->channelDimension.SetValue(pConfig->channelName.value);
 
+    this->labelDimension.SetName("Label");
```
disa6302, line 16: Just to be consistent, can we rename this to `WebRTCSDKCanaryType` or `WebRTCSDKCanaryLabel`?

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -29,8 +32,15 @@ VOID CloudwatchMonitoring::deinit()
 VOID CloudwatchMonitoring::push(const MetricDatum& datum)
 {
     Aws::CloudWatch::Model::PutMetricDataRequest cwRequest;
+    MetricDatum single = datum;
+    MetricDatum aggregated = datum;
+
+    single.AddDimensions(this->channelDimension);
+    aggregated.AddDimensions(this->labelDimension);
```
hassanctech, line 39: I like this idea but perhaps it would be better to add the label dimension to the single datum as well.  This way the label information available is while looking based on the single datum and if needed you can correlate that to the aggregated datum items.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
```diff
@@ -13,6 +13,9 @@ STATUS CloudwatchMonitoring::init()
     this->channelDimension.SetName("Channel");
```
disa6302, line 13: I know it is not part of this PR, but, to be more specific, can we change this to `WebRTCSDKCanaryChannelName`?

webrtc-c/canary/src/CloudwatchMonitoring.cpp
lherman-cs: Yep, that makes sense.

disa6302 (commented):
Minor comment. Rest LGTM.

webrtc-c/canary/src/CloudwatchMonitoring.cpp
lherman-cs: Updated to `WebRTCSDKCanaryLabel`

webrtc-c/canary/src/CloudwatchMonitoring.cpp
lherman-cs: Updated to `WebRTCSDKCanaryChannelName`

disa6302 (approved):

hassanctech (approved):
:)

PR #86, opened 2020-10-07
Add scenario label to Jenkins scripts

lherman-cs:
Currently, scenario label is the same with runner label. In the future, we will have more different sets of instances that are running with the same scenario. Scenario label is meant to be used to aggregate metrics at higher level.


webrtc-c/canary/jobs/runner.groovy
```diff
@@ -113,6 +115,7 @@ pipeline {
         string(name: 'MASTER_NODE_LABEL')
         string(name: 'VIEWER_NODE_LABEL')
         string(name: 'RUNNER_LABEL')
+        string(name: 'SCENARIO_LABEL')
```
disa6302, line 118: Do we need a new label? Why not use `RUNNER_LABEL` itself?

lherman-cs: `RUNNER_LABEL` is more granular and it's being used for the channel name. e.g. RUNNER_LABEL can be "WebrtcPeriodicLowBitrateEC2" and `SCENARIO_LABEL` can be "WebrtcPeriodic"

disa6302 (approved):

PR #98, opened 2020-10-12
Fix channel creation got stucked

lherman-cs:
Changes:
  * Add a small delay before running the viewer to avoid race condition
    that leads to getting stucked in channel creation
  * Fix typo in viewer's configuration


MushMal (approved):
Should we need a similar fix in the signaling canary?

lherman-cs:
> Should we need a similar fix in the signaling canary?

No need because in signaling canary, we ran them synchronously.

PR #110, opened 2020-10-16
Fix potential race condition in logger causing invalid log stream timestamps

disa6302:
*Issue #, if available:*

*Description of changes:*
Some logs might not get pushed if the log vector does not contain events in the appropriate timestamp order that might be caused due to a race condition. When different threads try to push logs into the events vector, a mutex would help ensuring a group of events are pushed into the vector in the appropriate order.

### DIFF
--- a/producer-c/producer-cloudwatch-integ/canary/CanaryLogsUtils.cpp
+++ b/producer-c/producer-cloudwatch-integ/canary/CanaryLogsUtils.cpp
@@ -5,6 +5,7 @@
 #include "CanaryUtils.h"
 
 PCloudwatchLogsObject gCloudwatchLogsObject = NULL;
+std::mutex gLogLock; // Protect access to gCloudwatchLogsObject
 
 STATUS initializeCloudwatchLogger(PCloudwatchLogsObject pCloudwatchLogsObject)
 {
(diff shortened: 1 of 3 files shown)


producer-c/producer-cloudwatch-integ/jobs/orchestrator.groovy
```diff
@@ -3,7 +3,7 @@ import jenkins.model.*
 WORKSPACE_PRODUCER="producer-c/producer-cloudwatch-integ"
 WORKSPACE_CONSUMER="consumer-java/aws-kinesis-video-producer-sdk-canary-consumer"
 GIT_URL='https://github.com/aws-samples/amazon-kinesis-video-streams-demos.git'
-GIT_HASH='master'
+GIT_HASH='fix-core-dump'
```
disa6302, line 6: Will change this back to master once the PR is approved

producer-c/producer-cloudwatch-integ/canary/CanaryLogsUtils.cpp
```diff
@@ -26,10 +26,13 @@ STATUS initializeCloudwatchLogger(PCloudwatchLogsObject pCloudwatchLogsObject)
 
 VOID setUpLogEventVector(PCHAR logString)
 {
-    Aws::String awsCwString((Aws::String) logString);
-    auto logEvent =
-        Aws::CloudWatchLogs::Model::InputLogEvent().WithMessage(awsCwString).WithTimestamp(GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
-    gCloudwatchLogsObject->canaryInputLogEventVec.push_back(logEvent);
+    if(gCloudwatchLogsObject != NULL) {
```
MushMal, line 29: spacing seems to be off

MushMal, line 29: You should grab the lock prior to the check

disa6302: The lock belong to the object. If the object is NULL, this would segfault
Hence, the lock is acquired inside.

disa6302: Oh. Seems fine locally. Let me re-check

MushMal: The issue is that if this is accessed by two parallel threads, both could proceed with checking the gCloudwatchLogsObject and for both of the threads the object can be NULL so you will still have a condition of double init. 
If this function is the only place the initialization of the object happens and two threads can access it, you need to sequentialize the access. The easiest is to define a scoped lock and interlock the check as well

disa6302: That makes sense. Let me do that.

producer-c/producer-cloudwatch-integ/canary/CanaryLogsUtils.cpp
```diff
@@ -5,6 +5,7 @@
 #include "CanaryUtils.h"
 
 PCloudwatchLogsObject gCloudwatchLogsObject = NULL;
+std::mutex logLock; // Protect access to gCloudwatchLogsObject
```
MushMal, line 8: I really don't like globals here at all. Is there a way to avoid them?

MushMal: gLogLock?

MushMal (approved):

producer-c/producer-cloudwatch-integ/canary/CanaryLogsUtils.cpp
disa6302: Not a fan of the global either. But for the current design, this is the only way. I am looking to revamp the canary at some point to move to a class based implementation. That will avoid use of globals altogether

PR #111, opened 2020-10-18
Reduce number of log files kept to reduce memory usage

disa6302:
*Issue #, if available:*

*Description of changes:*
Since the canaries push logs to cloudwatch, it is not necessary to maintain old logs for a long time since for long runs, logs bloat the memory usage on Jenkins leader. This PR reduces number of logs to be maintained to 1 and keeps it for 2 days only to provide a little relaxation and flexibility to check failed job logs within 2 days.

### DIFF
--- a/producer-c/producer-cloudwatch-integ/jobs/canary_seed.groovy
+++ b/producer-c/producer-cloudwatch-integ/jobs/canary_seed.groovy
@@ -5,8 +5,8 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.*;
 
 WORKSPACE="producer-c/producer-cloudwatch-integ"
 GROOVY_SCRIPT_DIR="$WORKSPACE/jobs"
-DAYS_TO_KEEP_LOGS=7
-NUMBER_OF_LOGS=5
+DAYS_TO_KEEP_LOGS=2
+NUMBER_OF_LOGS=1
 MAX_EXECUTION_PER_NODE=1
 NAMESPACE="producer"
 


MushMal (approved):
I understand it but why s it expensive to keep in both places? Some redundancy is ok. Ideally, we remove the logs after they have been successfully uploaded

disa6302:
> I understand it but why s it expensive to keep in both places? Some redundancy is ok. Ideally, we remove the logs after they have been successfully uploaded

When we **long** run these applications, the memory usage of the Jenkins leader shoots up because it saves the build history and we run out of memory within days. Ideally, if anything should go wrong, the expectation is to check what went wrong asap. If we want logs in 2 places, we can enable file logging in the application to allow logs to also be saved in the node the app is running. But, jenkins leader should not have a huge history kept over a week or two.

MushMal:
> > I understand it but why s it expensive to keep in both places? Some redundancy is ok. Ideally, we remove the logs after they have been successfully uploaded
> 
> When we **long** run these applications, the memory usage of the Jenkins leader shoots up because it saves the build history and we run out of memory within days. Ideally, if anything should go wrong, the expectation is to check what went wrong asap. If we want logs in 2 places, we can enable file logging in the application to allow logs to also be saved in the node the app is running. But, jenkins leader should not have a huge history kept over a week or two.
> There is redundancy, just that we are reducing the depth of redundancy by tracking only most recent logs.

I would imagine that the file logger is the right thing to do. Let’s go with this change and cut an issue to move to file logger. It might mean having a custom logger for webrtc to push the logs to both file and cw

PR #113, opened 2020-10-19
Unify webrtc and producer canaries

lherman-cs:
As of now, webrtc and producer live in separate projects. By unifying the two canaries, we will be able to deduplicate some of the codes and docs.


disa6302 (approved):
LGTM!

PR #116, opened 2020-10-21
Add configurable endpoint

lherman-cs:
By having the endpoint to be configurable, we'll be able to use development stages during use cases that require high TPS, e.g. load testing


canary/webrtc-c/src/Config.cpp
```diff
@@ -179,6 +179,7 @@ STATUS Config::initWithEnvVars()
         logLevel.initialized = TRUE;
     }
 
+    CHK_STATUS(optenv(CANARY_ENDPOINT_ENV_VAR, &endpoint, ""));
```
hassanctech, line 182: Should we specify a default for this?

lherman-cs: It's defined in the SDK when the string is empty or null:

 https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/2541b60c10753dbff8f35b3b3b528cf817b280f4/src/source/Signaling/ChannelInfo.c#L126-L132

disa6302 (approved):

MushMal (approved):

Issue #30, opened 2020-08-24
Improvements in producer C canary


disa6302:
- [x] Logging error handling - erro handle while creating LogStream
- [x] Adding support to display error number in application layer
- [x] Start up latency measurement
- [x] Set up permutation matrix
- [x] Have timestamp based cloudwatch logs
- [x] Have time based canary execution
- [x] Metric datum handling
- [x] Add clang format
- [x] Improve build system
- [x] Set up environment variables 
- [x] Add error rate tracking

PR #119, opened 2020-10-29
Start up latency for producer SDK

disa6302:
*Issue #, if available:*
#30 

*Description of changes:*
- Calculate start up latency in producer SDK
- Add the equivalent test suite to measure start up latency every minute into Jenkins pipeline

### DIFF
--- a/canary/producer-c/canary/CanaryStreamUtils.cpp
+++ b/canary/producer-c/canary/CanaryStreamUtils.cpp
@@ -297,6 +297,22 @@ CleanUp:
     return retStatus;
 }
 
+STATUS pushStartUpLatency(PCanaryStreamCallbacks pCanaryStreamCallbacks, DOUBLE startUpLatency)
+{
+    Aws::CloudWatch::Model::MetricDatum startupLatencyDatum, aggstartupLatencyDatum;
+    STATUS retStatus = STATUS_SUCCESS;
+    CHK(pCanaryStreamCallbacks != NULL, STATUS_NULL_ARG);
+    startupLatencyDatum.SetMetricName("StartupLatency");
+    startupLatencyDatum.AddDimensions(pCanaryStreamCallbacks->dimensionPerStream);
+    pushMetric(pCanaryStreamCallbacks, startupLatencyDatum, Aws::CloudWatch::Model::StandardUnit::Milliseconds, startUpLatency);
+
+    aggstartupLatencyDatum.SetMetricName("StartupLatency");
+    aggstartupLatencyDatum.AddDimensions(pCanaryStreamCallbacks->aggregatedDimension);
+    pushMetric(pCanaryStreamCallbacks, aggstartupLatencyDatum, Aws::CloudWatch::Model::StandardUnit::Milliseconds, startUpLatency);
+CleanUp:
+    return STATUS_SUCCESS;
+}
+
 STATUS computeStreamMetricsFromCanary(STREAM_HANDLE streamHandle, PCanaryStreamCallbacks pCanaryStreamCallbacks)
 {
     STATUS retStatus = STATUS_SUCCESS;
--- a/canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
+++ b/canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
@@ -197,8 +197,11 @@ INT32 main(INT32 argc, CHAR* argv[])
     BOOL cleanUpDone = FALSE;
     BOOL fileLoggingEnabled = FALSE;
     BOOL runTillStopped = FALSE;
-
+    PAuthCallbacks pAuthCallbacks = NULL;
     CanaryConfig config;
+    BOOL firstFrame = TRUE;
+    UINT64 startTime;
+    DOUBLE startUpLatency;
     initializeEndianness();
     SRAND(time(0));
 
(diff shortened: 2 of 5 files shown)


canary/producer-c/canary/CanaryStreamUtils.cpp
```diff
@@ -297,6 +297,21 @@ STATUS publishErrorRate(STREAM_HANDLE streamHandle, PCanaryStreamCallbacks pCana
...
+    Aws::CloudWatch::Model::MetricDatum startupLatencyDatum, aggstartupLatencyDatum;
+    startupLatencyDatum.SetMetricName("StartupLatency");
+    startupLatencyDatum.AddDimensions(pCanaryStreamCallbacks->dimensionPerStream);
+    pushMetric(pCanaryStreamCallbacks, startupLatencyDatum, Aws::CloudWatch::Model::StandardUnit::Milliseconds, pCanaryStreamCallbacks->startUpLatency);
+
+    aggstartupLatencyDatum.SetMetricName("StartupLatency");
+    aggstartupLatencyDatum.AddDimensions(pCanaryStreamCallbacks->aggregatedDimension);
+    pushMetric(pCanaryStreamCallbacks, aggstartupLatencyDatum, Aws::CloudWatch::Model::StandardUnit::Milliseconds, pCanaryStreamCallbacks->startUpLatency);
+
+CleanUp:
```
MushMal, line 311: Not sure if you need this label. It will generate warnings if not used. Either remove this or if the input is not sanitized add CHK(pCanaryStreamCallbacks != NULL, STATUS_BLAH);

canary/producer-c/canary/CanaryStreamUtils.cpp
```diff
@@ -297,6 +297,21 @@ STATUS publishErrorRate(STREAM_HANDLE streamHandle, PCanaryStreamCallbacks pCana
     return retStatus;
 }
 
+STATUS pushStartUpLatency(PCanaryStreamCallbacks pCanaryStreamCallbacks)
```
MushMal, line 300: Just for my understanding.. We push the exact same value to two metrics? One that's aggregated and one that's not?

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
```diff
@@ -264,8 +266,25 @@ INT32 main(INT32 argc, CHAR* argv[])
         // adjust members of pStreamInfo here if needed
         pStreamInfo->streamCaps.nalAdaptationFlags = NAL_ADAPTATION_FLAG_NONE;
 
-        CHK_STATUS(createDefaultCallbacksProviderWithAwsCredentials(accessKey, secretKey, sessionToken, MAX_UINT64, region, cacertPath, NULL, NULL,
-                                                                    &pClientCallbacks));
+        startTime = GETTIME();
+        CHK_STATUS(createAbstractDefaultCallbacksProvider(DEFAULT_CALLBACK_CHAIN_COUNT,
+                                                          API_CALL_CACHE_TYPE_NONE,
```
MushMal, line 271: I see. Cool.

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
```diff
@@ -330,7 +349,12 @@ INT32 main(INT32 argc, CHAR* argv[])
                 lastKeyFrameTimestamp = frame.presentationTs;
             }
             CHK_STATUS(putKinesisVideoFrame(streamHandle, &frame));
-
+            if(firstFrame) {
```
MushMal, line 352: spacing after if (

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
```diff
@@ -197,8 +197,10 @@ INT32 main(INT32 argc, CHAR* argv[])
     BOOL cleanUpDone = FALSE;
     BOOL fileLoggingEnabled = FALSE;
     BOOL runTillStopped = FALSE;
-
+    PAuthCallbacks pAuthCallbacks = NULL;
     CanaryConfig config;
+    BOOL firstFrame = TRUE;
+    UINT64 startTime = 0;
```
MushMal, line 203: Do we need to init startTime here? Some static analysis tools will generate warnings for this as its value gets overwritten later with GETTIME(). It's a minor issue so feel free to ignore.

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
MushMal: Could you add some comments about firstFrame? Why do we measure it here instead of before the loop? I mean, we do create in a sync fashion right? What's the delta between outputting here and right after the createsync succeeded? There will be a few hundreds of ns correct?

canary/producer-c/jobs/orchestrator.groovy
```diff
@@ -129,6 +129,19 @@ pipeline {
                             ],
                             wait: false
                         )
+                        build(
```
MushMal, line 132: Can we get @lherman-cs to look into this as well?

MushMal (commented):
Looks good but some comments to understand why are we measuring latency after the first putFrame. The only thing that comes to my mind is the fact that the first putFrame will drive the state machinery asynchronously to issue PutMedia call. If this is the case, it would be nice to add it to the comments. NOTE: The PutMedia call that's issued with the first frame is an async call - all together I would imagine it would take a few hundreds of nanos or a couple of microseconds measured after the first putFrame or after the stream is created in a sync fashion.

canary/producer-c/canary/CanaryStreamUtils.cpp
disa6302: Yes.

canary/producer-c/canary/CanaryStreamUtils.cpp
disa6302: Makes sense. My bad. Will remove it.

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
disa6302: I usually have the habit of assigning an initial value to a variable. I can remove this.

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
```diff
@@ -330,7 +350,12 @@ INT32 main(INT32 argc, CHAR* argv[])
                 lastKeyFrameTimestamp = frame.presentationTs;
             }
             CHK_STATUS(putKinesisVideoFrame(streamHandle, &frame));
-
+            startUpLatency = (DOUBLE)(GETTIME() - startTime) / (DOUBLE) HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
```
MushMal: Move this into the if loop as it's only used when publishing the metrics. No need to calculate it for no reason.

MushMal (commented):
Minor comment

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
disa6302: That is correct. Added a comment to explain why it is measured here

MushMal (approved):

lherman-cs (approved):
LGTM

PR #122, opened 2020-11-06
Use EMA instead of vector for e2e metric

lherman-cs:

### DIFF
--- a/canary/webrtc-c/src/CloudwatchMonitoring.cpp
+++ b/canary/webrtc-c/src/CloudwatchMonitoring.cpp
@@ -233,26 +233,28 @@ VOID CloudwatchMonitoring::pushInboundRtpStats(Canary::PIncomingRTPMetricsContex
     this->push(incomingFrameDropRateDatum);
 }
 
-VOID CloudwatchMonitoring::pushEndToEndMetrics(Canary::PEndToEndMetricsContext pEndToEndMetricsContext)
+VOID CloudwatchMonitoring::pushEndToEndMetrics(Canary::EndToEndMetricsContext ctx)
 {
     MetricDatum endToEndLatencyDatum, sizeMatchDatum;
+    DOUBLE latency = ctx.frameLatencyAvg / (DOUBLE) HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
 
+    // TODO: due to https://github.com/aws-samples/amazon-kinesis-video-streams-demos/issues/96,
+    //       it's not clear why the emitted metric shows -nan. Since -nan is a string that's outputted
+    //       from Datum's value stringify implementation, we should try to get the original value by
+    //       printing it ourself.
+    //
+    //       If the issues doesn't exist anymore, please remove this as this is intended for debugging only.
+    //       The generic metric logging should be sufficient.
+    DLOGD("Current end-to-end frame latency: %4.2lf", latency);
     endToEndLatencyDatum.SetMetricName("EndToEndFrameLatency");
     endToEndLatencyDatum.SetUnit(StandardUnit::Milliseconds);
-    auto& latencies = pEndToEndMetricsContext->frameLatency;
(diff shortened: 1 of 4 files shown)


disa6302:
How does changing to sum and count help?

lherman-cs:
@disa6302 It doesn't it's just an enhancement. But, the lock here https://github.com/aws-samples/amazon-kinesis-video-streams-demos/pull/122/files#diff-36c06b4e0a11b29790dda910ed5a0beb1cf833082548e8ebe1370d8be1c8e476R703, will help with the race condition.

canary/webrtc-c/src/CloudwatchMonitoring.cpp
```diff
@@ -233,26 +233,28 @@ VOID CloudwatchMonitoring::pushInboundRtpStats(Canary::PIncomingRTPMetricsContex
...
+    DOUBLE latency = ctx.frameLatencyAvg / (DOUBLE) HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+
+    // TODO: due to https://github.com/aws-samples/amazon-kinesis-video-streams-demos/issues/96,
+    //       it's not clear why the emitted metric shows -nan. Since -nan is a string that's outputted
+    //       from Datum's value stringify implementation, we should try to get the original value by
+    //       printing it ourself.
+    //
+    //       If the issues doesn't exist anymore, please remove this as this is intended for debugging only.
+    //       The generic metric logging should be sufficient.
+    DLOGD("Current end-to-end frame latency: %4.2lf", latency);
```
disa6302, line 248: This is logged in the common functionality right? We will still be logging the aggregate here.

MushMal (approved):

canary/webrtc-c/src/Peer.cpp
```diff
@@ -536,11 +536,16 @@ STATUS Peer::addTransceiver(RtcMediaStreamTrack& track)
         frameDataPtr += SIZEOF(UINT64);
         UINT32 receivedSize = getUnalignedInt32BigEndian((PINT32)(frameDataPtr));
 
-        pPeer->endToEndMetricsContext.frameLatency.push_back((DOUBLE)(GETTIME() - receivedTs) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+        pPeer->endToEndMetricsContext.frameLatencyAvg =
+            EMA_ACCUMULATOR_GET_NEXT(pPeer->endToEndMetricsContext.frameLatencyAvg, GETTIME() - receivedTs);
```
disa6302, line 540: Just for my understanding. What is the hypothesis for EMA fixing the e2e metric emitting problem?

hassanctech (approved):

canary/webrtc-c/src/CloudwatchMonitoring.cpp
lherman-cs: It's as what the comment says, the common logging uses C++ string, which is entirely dependent on the implementor. What we want is the original value.

canary/webrtc-c/src/Peer.cpp
lherman-cs: Again, it's not the EMA or how it's calculated, but there's a data race in collecting and publishing these metrics. Therefore, I added this lock, https://github.com/aws-samples/amazon-kinesis-video-streams-demos/pull/122/files#diff-36c06b4e0a11b29790dda910ed5a0beb1cf833082548e8ebe1370d8be1c8e476R702.

EMA is an enhancement from what we had, and should be preferred as it's standard in our SDKs to calculate average in a stream.

disa6302: Ok. Makes sense.

canary/webrtc-c/src/CloudwatchMonitoring.cpp
disa6302: Ah! Ok. Now I get it.

disa6302 (approved):

Issue #50, opened 2020-09-18
Start up latency in webrtc canary


disa6302:
We need to record the start up latency in the canary and push as a metric

PR #124, opened 2020-11-19
Start up latency for frame write

disa6302:
*Issue #, if available:*
#50 
*Description of changes:*
- Add latency from receiving offer to sending out first frame

### DIFF
--- a/canary/webrtc-c/src/CloudwatchMonitoring.cpp
+++ b/canary/webrtc-c/src/CloudwatchMonitoring.cpp
@@ -166,6 +166,16 @@ VOID CloudwatchMonitoring::pushSignalingRoundtripLatency(UINT64 delay, StandardU
     this->push(datum);
 }
 
+VOID CloudwatchMonitoring::pushTimeToFirstFrame(UINT64 timeToFirstFrame, StandardUnit unit)
+{
+    MetricDatum datum;
+
+    datum.SetMetricName("TimeToFirstFrame");
+    datum.SetValue(timeToFirstFrame);
+    datum.SetUnit(unit);
+
+    this->push(datum);
+}
 VOID CloudwatchMonitoring::pushSignalingInitDelay(UINT64 delay, StandardUnit unit)
 {
     MetricDatum datum;
--- a/canary/webrtc-c/src/Peer.cpp
+++ b/canary/webrtc-c/src/Peer.cpp
@@ -31,6 +31,7 @@ STATUS Peer::init(const Canary::PConfig pConfig, const Callbacks& callbacks)
     this->canaryIncomingRTPMetricsContext.prevBytesReceived = 0;
     this->canaryIncomingRTPMetricsContext.prevFramesDropped = 0;
     this->canaryIncomingRTPMetricsContext.prevTs = GETTIME();
+    this->firstFrame = TRUE;
 
     CHK_STATUS(createStaticCredentialProvider((PCHAR) pConfig->accessKey.value.c_str(), 0, (PCHAR) pConfig->secretKey.value.c_str(), 0,
                                               (PCHAR) pConfig->sessionToken.value.c_str(), 0, MAX_UINT64, &pAwsCredentialProvider));
@@ -415,7 +416,7 @@ STATUS Peer::handleSignalingMsg(PReceivedSignalingMessage pMsg)
         RtcSessionDescriptionInit offerSDPInit, answerSDPInit;
         NullableBool canTrickle;
         UINT32 buffLen;
-
+        this->offerReceiveTimestamp = GETTIME();
         if (!this->isMaster) {
             DLOGW("Unexpected message SIGNALING_MESSAGE_TYPE_OFFER");
             CHK(FALSE, retStatus);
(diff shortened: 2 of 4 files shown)


canary/webrtc-c/src/Peer.cpp
```diff
@@ -602,6 +603,13 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
             retStatus = STATUS_SUCCESS;
         }
 
+        else if (STATUS_SUCCEEDED(retStatus) && this->firstFrame && this->isMaster) {
```
MushMal, line 606: No need to do else if - just do if

canary/webrtc-c/src/Peer.cpp
```diff
@@ -602,6 +603,13 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
...
+            this->firstFrame = FALSE;
+            this->startupLatency = DOUBLE(GETTIME() - this->offerReceiveTimestamp) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
```
MushMal, line 608: This is a bad construct. Perhaps what you meant is this?

this->startupLatency = (DOUBLE) (GETTIME() - this->offerReceiveTimestamp) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;

canary/webrtc-c/src/Peer.h
```diff
@@ -84,6 +84,9 @@ class Peer {
     std::vector<PRtcRtpTransceiver> videoTransceivers;
     BOOL isMaster;
     BOOL trickleIce;
+    UINT64 offerReceiveTimestamp;
+    BOOL firstFrame;
+    DOUBLE startupLatency;
```
MushMal, line 89: This can be a local variable

canary/webrtc-c/src/CloudwatchMonitoring.cpp
```diff
@@ -166,6 +166,16 @@ VOID CloudwatchMonitoring::pushSignalingRoundtripLatency(UINT64 delay, StandardU
     this->push(datum);
 }
 
+VOID CloudwatchMonitoring::pushStartUpLatency(UINT64 latency, StandardUnit unit)
+{
+    MetricDatum datum;
+
+    datum.SetMetricName("StartUpLatency");
```
MushMal, line 173: Should this be named as "TimeToFirstFrame"? or something similar? Startup latency sounds a little vague

canary/webrtc-c/src/Peer.cpp
disa6302: It will have to be else if because we set `retStatus` to `STATUS_SUCCESS` in the previous if call. I can reorder the ifs, but this should be okay.

canary/webrtc-c/src/Peer.cpp
disa6302: Uh-oh! Typo :(

canary/webrtc-c/src/Peer.cpp
MushMal: Oh, of course. I am blind - I thought it simply returns the value there.
 
The suggestion I would have would be something like this:

retStatus = func();

CHK (retStatus == WHAT_TO_CHECK || retStatus == STATUS_SUCCESS, retStatus);

// Reset the retStatus
retStatus = STATUS_SUCCESS;

if (this->firstFrame && blah) {...

canary/webrtc-c/src/Peer.h
MushMal, line 89: Remove this?

MushMal (commented):
Minor comment about removing startupLatency and I also commented for the return check and if/else code. This is a more of a choice comment so you can simply ignore it.

canary/webrtc-c/src/Peer.cpp
disa6302: Ok. This makes it cleaner. Changed.

disa6302:
> Minor comment about removing startupLatency and I also commented for the return check and if/else code. This is a more of a choice comment so you can simply ignore it.

Fixed it. Makes the code cleaner.

canary/webrtc-c/src/Peer.cpp
```diff
@@ -597,17 +598,18 @@ STATUS Peer::writeFrame(PFrame pFrame, MEDIA_STREAM_TRACK_KIND kind)
     }
     for (auto& transceiver : transceivers) {
         retStatus = ::writeFrame(transceiver, pFrame);
-        if (retStatus == STATUS_SRTP_NOT_READY_YET) {
-            // do nothing
-            retStatus = STATUS_SUCCESS;
+        CHK (retStatus == STATUS_SRTP_NOT_READY_YET || retStatus == STATUS_SUCCESS, retStatus);
+
+        if (this->firstFrame && this->isMaster) {
```
MushMal, line 603: Oh, we might need to think about this in case of retStatus == STATUS_SRTP_NOT_READY_YET. Perhaps we need to do the differentiation here and reset the value latest before exiting.

if (this->firstFrame && this->isMaster) {
...
}
retStatus = STATUS_SUCCESS; // Reset the status

disa6302: Oops! My bad. Completely missed the case. I fixed it now.

MushMal (approved):

PR #137, opened 2021-01-15
Fix high latency shoot ups for intermittent producer canary

### LINKED ISSUE #136: Persisted Ack Latency and Received Ack latency shoots up to timestamp values when stopping streaming
While running the canaries it was observed that while running the intermittent scenario as:

```
while(time < runTill) {
    // putFrame
}
while(time < sleepTill) {
    // sleep
}

Right when it begins to sleep, the persisted Ack latency shoots up to values as high as a timestamp when the max expected latency by design is about 20 - 25 seconds tops.

disa6302:
*Issue #, if available:*
#136 

*Description of changes:*
- Previously, we recorded end fragment time only on every key frame. However, in intermittent scenario, it is possible we stop the `putFrame` operation on a non key frame which leads to the fragment's key frame timestamp to be added to the map after the sleep time. Since the SDK sends a persisted Ack after closing the fragment, the map might not have the recorded key frame time when the ack arrives, leading to a random value being used [here](https://github.com/aws-samples/amazon-kinesis-video-streams-demos/blob/9559671c83a14a9247e479f516b50c6018af427b/canary/producer-c/canary/CanaryStreamUtils.cpp#L184), that inturn leads to bizarre values for Ack latency. This change fixes it by recording the key frame timestamp right before `sleep`.

- This PR also updates producer C to most recent commit.

Resolves #136 

### DIFF
--- a/canary/producer-c/CMake/Dependencies/libkvsProducerC-CMakeLists.txt
+++ b/canary/producer-c/CMake/Dependencies/libkvsProducerC-CMakeLists.txt
@@ -6,7 +6,7 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsProducerC-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
-    GIT_TAG           17d875fc54e7378f1c1483854990582fb7aa4336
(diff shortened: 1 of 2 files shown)


hassanctech (approved):
lgtm!

Issue #144, opened 2021-02-08
Add multi track for intermittent producer canary


disa6302:
The canaries needs to be set up with at least 2 tracks to properly test the auto intermittent producer scenario and ensure the options are set up appropriately for streaming.

disa6302:
Change merged. Resolving.

PR #145, opened 2021-02-09
Intermittent producer in multitrack

disa6302:
*Issue #, if available:*
#144 
*Description of changes:*
- Add multi track option for producer canary
- Include a job to run intermittent scenario in single track and multi track scenario
- Update commit hash to include latest fix

### DIFF
--- a/canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
+++ b/canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
@@ -97,6 +97,8 @@ STATUS parseConfigFile(PCanaryConfig pCanaryConfig, PCHAR filePath)
             i++;
         } else if (compareJsonString((PCHAR) params, &tokens[i], JSMN_STRING, CANARY_SCENARIO_ENV_VAR)) {
             getJsonValue(params, tokens[i + 1], pCanaryConfig->canaryScenario);
+        } else if (compareJsonString((PCHAR) params, &tokens[i], JSMN_STRING, CANARY_TRACK_TYPE_ENV_VAR)) {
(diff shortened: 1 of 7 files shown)


canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
```diff
@@ -366,7 +378,16 @@ INT32 main(INT32 argc, CHAR* argv[])
...
             if (GETTIME() < runTill) {
+                frame.trackId = DEFAULT_VIDEO_TRACK_ID;
                 CHK_STATUS(putKinesisVideoFrame(streamHandle, &frame));
+
+                // Send frame on another track only if we want to run multi track. For the sake of
+                // multitrack, we use the same frame for video and audio and just modify the flags.
+                if (STRCMP(config.canaryTrackType, CANARY_MULTI_TRACK_TYPE) == 0) {
+                    frame.flags = FRAME_FLAG_NONE;
+                    frame.trackId = DEFAULT_AUDIO_TRACK_ID;
+                    CHK_STATUS(putKinesisVideoFrame(streamHandle, &frame));
```
MushMal, line 389: Question. What happens if we get an error on putframe (or other API calls for that matter here). Do we end up just logging it or we have a way to output error metrics? I presume the latter but wanted to confirm.

MushMal (approved):
Nice!

canary/producer-c/canary/KvsProducerSampleCloudwatch.cpp
disa6302: You are right. PutFrame error metrics are emitted in the canary to cloudwatch.

hassanctech (approved):
Great!

PR #148, opened 2021-02-22
Ensure instructions point to opening a page only in localhost

disa6302:
*Issue #, if available:*

*Description of changes:*

Update the instruction to be precise to open webpage on local host only.


MushMal (approved):
hmm ok

PR #150, opened 2021-02-25
Initial work on KVS GST Plugin

MushMal:
*Issue #, if available:*

*Description of changes:*


hassanctech (approved):
I know this is initial commit, looks good, but we should add lots of testing around methods like:

`identifyFrameNalFormat`
`identifyCpdNalFormat `
`convertCpdFromAvcToAnnexB `
`convertCpdFromHevcToAnnexB `

disa6302:
It looks good. Though this is an initial commit, can we have a README with some details on what this demo is intended to do? We might ultimately move this to another place, but having some details would be useful either ways.

disa6302 (approved):

Issue #152, opened 2021-03-04
[GST Plugin] Signaling channel not respecting `aws-region` property


adamthesax:
Hi! I am working on getting the GST plugin up and running so I can push video to both a video stream and signaling channel. While the plugin is working in the sense that it is publishing to both a video stream an signaling channel, the signaling channel is not respecting the `aws-region` property on `kvsplugin` and is always pushing to `us-west-2`

The problem is very reproducible with any command, but for reference this is the test command I am running:
```
gst-launch-1.0 uridecodebin uri=https://ia800701.us.archive.org/26/items/SampleVideo1280x7205mb/SampleVideo_1280x720_5mb.mp4 ! vtenc_h264_hw max-keyframe-interval=30 bitrate=500 ! kvsplugin aws-region=us-east-1
```

And this is how I am verifying the signaling channel is being created in the incorrect region:
```
$ aws kinesisvideo list-streams --region us-east-1
{
    "StreamInfoList": [
        {
            "DeviceName": "VPI9Y8SDEEW3OTOY",
            "StreamName": "DEFAULT_STREAM",
            "StreamARN": "arn:aws:kinesisvideo:us-east-1:XXXXXXXXXXXX:stream/DEFAULT_STREAM/1614895586442",
            "MediaType": "video/h264",
            "KmsKeyId": "arn:aws:kms:us-east-1:028923058570:alias/aws/kinesisvideo",
            "Version": "8i2cifJ2iTKCh4zHfOuO",
            "Status": "ACTIVE",
            "CreationTime": "2021-03-04T17:06:26.442000-05:00",
            "DataRetentionInHours": 2
        }
    ]
}

$ aws kinesisvideo list-signaling-channels --region us-east-1
{
    "ChannelInfoList": []
}

$ aws kinesisvideo list-signaling-channels --region us-west-2
{
    "ChannelInfoList": [
        {
            "ChannelName": "DEFAULT_CHANNEL",
            "ChannelARN": "arn:aws:kinesisvideo:us-west-2:XXXXXXXXXXXX:channel/DEFAULT_CHANNEL/1614895587705",
            "ChannelType": "SINGLE_MASTER",
            "ChannelStatus": "ACTIVE",
            "CreationTime": "2021-03-04T17:06:27.705000-05:00",
            "SingleMasterConfiguration": {
                "MessageTtlSeconds": 60
            },
            "Version": "IEoqKrmPSLsMvE18zgjT"
        }
[...]

PR #156, opened 2021-03-12
Pass configured region to WebRTC lib in GST plugin

adamthesax:
*Issue #, if available: #152*

*Description of changes:*
This passes the region given to the GST plugin via the `aws-region` property to the WebRTC library. This ensure that when a region is specified to the GST plugin that both the Video Stream and Signaling Channel are created in the requested region.


disa6302:
Thank you for the changes @adamthesax !

disa6302 (approved):

PR #190, opened 2021-10-13
Backoff

disa6302:
*Issue #, if available:*

*Description of changes:*
Include exponential backoff usage in WebRTC demos


jdelapla:
in ~Peer() we need to add: 
`exponentialBackoffStateFree(&pExponentialBackoffState);`

However we also need to change the scope of **pExponentialBackoffState** so that the destructor can see it.

PR #217, opened 2021-12-20
Use correct API to get signaling retry count to publish to cloud watch

kateyanurag:
*Issue #, if available:*

*Description of changes:*
Use correct API to get signaling retry count to publish to cloud watch


disa6302 (commented):
Note that the retry count would not change if the signaling client creation fails in a loop and the application exits. This was the whole point of using the read only value because that way we would have the value throughout the life of the signaling client object.

Why are we making this change? Because the readOnly value is updated everytime we complete a call to the createSignalingClient/fetch(if I recall correctly) anyways.

If the create/fetch call fails after max retries, the signalingClientMetrics object would be Null and we would never get the metric out. (Basically, we would fail with STATUS_NULL_ARG in this function).

PR #249, opened 2022-07-19
Producer CPP Canary

stefankiesz:


canary/producer-cpp/src/CanaryLogs.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+VOID CanaryLogs::onPutLogEventResponseReceivedHandler(const Aws::CloudWatchLogs::CloudWatchLogsClient* cwClientLog,
+                                          const Aws::CloudWatchLogs::Model::PutLogEventsRequest& request,
+                                          const Aws::CloudWatchLogs::Model::PutLogEventsOutcome& outcome,
+                                          const std::shared_ptr<const Aws::Client::AsyncCallerContext>& context)
+{
+    if (!outcome.IsSuccess()) {
+        DLOGE("Failed to push logs: %s", outcome.GetError().GetMessage().c_str());
+    } else {
+        DLOGS("Successfully pushed logs to cloudwatch");
```
disa6302, line 55: Replace all these with `LOG_x`. DLOGs wont get pushed to cloudwatch

canary/producer-cpp/.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
disa6302: Remove file

canary/producer-cpp/CMake/Utilities.cmake
```diff
@@ -0,0 +1,54 @@
+# build library from source
```
disa6302: Remove file

canary/producer-cpp/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
+cmake_minimum_required(VERSION 3.16)
+# set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake;${CMAKE_MODULE_PATH}")
```
disa6302, line 2: Remove if not required

canary/producer-cpp/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+
+include(FetchContent)
+
+
+if(APPLE AND NOT DEFINED ENV{MACOSX_DEPLOYMENT_TARGET} AND NOT DEFINED ENV{SDKROOT})
+  set(CMAKE_OSX_DEPLOYMENT_TARGET ${CMAKE_SYSTEM_VERSION})
+endif()
+
+# include(Utilities)
```
disa6302, line 11: Remove if not required. Goes for all the commented out pieces.

canary/producer-cpp/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+project(KinesisVideoProducerCloudwatch LANGUAGES C CXX) # what to do here, take out C? why is cpp called CXX in this case?
+
+# set(CMAKE_MACOSX_RPATH TRUE)
+# get_filename_component(ROOT "${CMAKE_CURRENT_SOURCE_DIR}" ABSOLUTE)
+
+
+set(BUILD_GSTREAMER_PLUGIN ON CACHE BOOL "Tell SDK to ..." FORCE)
```
disa6302, line 18: Fix the message. "Tell SDK to..."

canary/producer-cpp/README.md
```diff
@@ -0,0 +1,2 @@
+Welcome to Kinesis Video Streams CPP Producer SDK integration with AWS Cloudwatch Metrics!
```
disa6302, line 1: Just adding a note: Needs to be completed.

canary/producer-cpp/jobs/canary_seed.groovy
```diff
@@ -0,0 +1,84 @@
...
+    "method jenkins.model.Jenkins getItemByFullName java.lang.String",
+    "method jenkins.model.ParameterizedJobMixIn\$ParameterizedJob isDisabled",
+    "method jenkins.model.ParameterizedJobMixIn\$ParameterizedJob setDisabled boolean",
+    "method org.jenkinsci.plugins.workflow.job.WorkflowRun doKill",
+    "staticMethod jenkins.model.Jenkins getInstance",
+    "staticField java.lang.Long MAX_VALUE"
+])
+
+pipelineJob("${NAMESPACE}-orchestrator") {
+	description('Run producer')
```
disa6302, line 39: producer cpp canary. Be specific. Same goes for runner job descriptions.

canary/producer-cpp/jobs/orchestrator.groovy
```diff
@@ -0,0 +1,217 @@
+import jenkins.model.*
+
+WORKSPACE_PRODUCER="src"
+// WORKSPACE_CONSUMER="consumer-java"
+GIT_URL='https://github.com/aws-samples/amazon-kinesis-video-streams-demos.git'
+GIT_HASH='new-canary-producer-cpp'
```
disa6302, line 6: Remember to change this to master before merging because the branch automatically gets deleted once the PR is merged.

canary/producer-cpp/jobs/orchestrator.groovy
```diff
@@ -0,0 +1,217 @@
...
+RUNNER_JOB_NAME_PREFIX = "producer-cpp-runner"
+
+// TODO: Set up configurability to run different parameter combinations
+// Run long run canary for 12 hours
+LONG_RUN_DURATION_IN_SECONDS = 12 * 60 * 60
+SHORT_RUN_DURATION_IN_SECONDS = 30
+COLD_STARTUP_DELAY_IN_SECONDS = 60 * 60
+MIN_RETRY_DELAY_IN_SECONDS = 60
+FRAGMENT_SIZE_IN_BYTES = 250000
```
disa6302, line 15: Is this needed for the canary?

canary/producer-cpp/jobs/orchestrator.groovy
```diff
@@ -0,0 +1,217 @@
...
+                        // TODO: Use matrix to provide configurability in parameters
+                        build(
+                            job: NEXT_AVAILABLE_RUNNER,
+                            parameters: COMMON_PARAMS + [
+                                string(name: 'CANARY_STREAM_NAME', value: "Continuous-Longrun-01"),
+                                string(name: 'CANARY_DURATION_IN_SECONDS', value: LONG_RUN_DURATION_IN_SECONDS.toString()),
+                                string(name: 'PRODUCER_NODE_LABEL', value: "producer-uw2"),
+                                // string(name: 'CANARY_TYPE', value: "RealtimeStatic"),
+                                string(name: 'RUNNER_LABEL', value: "Longrun"),
+                                // string(name: 'FRAGMENT_SIZE_IN_BYTES', value: FRAGMENT_SIZE_IN_BYTES.toString()),
```
disa6302, line 126: Remove commented out portions if not applicable please.

canary/producer-cpp/kvs_log_configuration
```diff
@@ -0,0 +1,16 @@
+log4cplus.rootLogger=DEBUG, KvsConsoleAppender, KvsFileAppender
```
disa6302, line 1: This file is not necessary since we end up copying as part of CMakeLists.txt file. Remove the file.

disa6302 (commented):
General feedback: Let us get rid of all the commented out portions of code in the files.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+        uint64_t timeOfFragmentEndSent = data->timeOfNextKeyFrame->find(pFragmentAck->timestamp)->second;
+        cout << "Map Debug: with value of " << timeOfFragmentEndSent << endl;
+        cout << "Map Debug: latency: " << duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count() - timeOfFragmentEndSent << endl;
+
+        // When Canary sleeps, timeOfFragmentEndSent become less than currentTimeStamp, don't send those metrics
+        if (timeOfFragmentEndSent > pFragmentAck->timestamp)
+        {
+            if (pFragmentAck->ackType == FRAGMENT_ACK_TYPE_PERSISTED)
+            {
+                Aws::CloudWatch::Model::MetricDatum persistedAckLatency_datum;
```
disa6302, line 171: General guidance: Let us following one style of naming: either camel case: `persistedAckLatencyDatum` or `persisted_ack_latency_datum`. Let us mix it. Since most of the code is using former, please modify relevant function names and variable names to be all camel cases type.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+                auto currentTimestamp = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
+                auto persistedAckLatency = (currentTimestamp - timeOfFragmentEndSent); // [milliseconds]
+                pushMetric("PersistedAckLatency", persistedAckLatency, Aws::CloudWatch::Model::StandardUnit::Milliseconds, persistedAckLatency_datum, data->pDimension_per_stream, cwRequest);
+                LOG_DEBUG("Persisted Ack Latency: " << persistedAckLatency);
+                if (data->pCanaryConfig->useAggMetrics)
+                {
+                    pushMetric("PersistedAckLatency", persistedAckLatency, Aws::CloudWatch::Model::StandardUnit::Milliseconds, persistedAckLatency_datum, data->pAggregated_dimension, cwRequest);
+
+                }
+                auto outcome = data->pCWclient->PutMetricData(cwRequest);
```
disa6302, line 184: Let us change to an async call. Use PutMetricDataAsync instead. We do not want to block on this. You can refer to producer C canary for reference.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+                    if (!outcome.IsSuccess())
+                    {
+                        cout << "Failed to put ReceivedAckLatency metric data:" << outcome.GetError().GetMessage() << endl;
+                    }
+                    else
+                    {
+                        cout << "Successfully put ReceivedAckLatency metric data" << endl;
+                    }
+                }
+        } else
```
disa6302, line 218: This else portion does not make sense. I think what you are going with if it neither received or persisted ack. However, we still want to track other acks (error and buffered( even if we are not going to publish metrics for them. It can be a simple log line, but we need it. So when you modify this to switch case, let us add error ack and buffered ack type as well.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+    pCurPtr += SIZEOF(UINT32);
+    putUnalignedInt32BigEndian((PINT32) pCurPtr, COMPUTE_CRC32(pFrame->frameData, pFrame->size));
+}
+
+VOID createCanaryFrameData(PFrame pFrame)
+{
+    UINT32 i;
+
+    for (i = CANARY_METADATA_SIZE; i < pFrame->size; i++) {
+        pFrame->frameData[i] = RAND();
```
disa6302, line 248: Why this? We dont need random data because we have our actual gstreamer source frame data, right? You need to append the metadata to the actual frame data.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+    frame->decodingTs = static_cast<UINT64>(dts.count()) / DEFAULT_TIME_UNIT_IN_NANOS;
+    frame->presentationTs = static_cast<UINT64>(pts.count()) / DEFAULT_TIME_UNIT_IN_NANOS;
+    // set duration to 0 due to potential high spew from rtsp streams
+    frame->duration = 0;
+    frame->size = static_cast<UINT32>(len);
+    frame->frameData = reinterpret_cast<PBYTE>(data);
+    frame->trackId = DEFAULT_TRACK_ID;
+}
+
+void updateFragmentEndTimes(UINT64 curKeyFrameTime, uint64_t &lastKeyFrameTime, map<uint64_t, uint64_t> *mapPtr)
```
disa6302, line 265: Hmm. Doesnt make sense. Why do you need an address for lastKeyFrameTime? Why are you using address instead of pointer? It needs to be `PUINT64 lastKeyFrameTime`. And in the place where this is being called, you need to send address.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+                }
+            }
+        }
+        lastKeyFrameTime = curKeyFrameTime;
+}
+
+void pushKeyFrameMetrics(Frame frame, CustomData *cusData)
+{
+    updateFragmentEndTimes(frame.presentationTs, cusData->lastKeyFrameTime, cusData->timeOfNextKeyFrame);
+
```
disa6302, line 291: Can we modularize this function further to get stream metrics and client metrics respectively? It will be cleaner.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+        pushMetric("ContentStoreAvailableSize", availableStoreSize, Aws::CloudWatch::Model::StandardUnit::Kilobytes, metricDatum, cusData->pAggregated_dimension, cwRequest);
+    }
+
+    // Capture error rate metrics every 60 seconds
+    double duration = duration_cast<seconds>(system_clock::now().time_since_epoch()).count() - cusData->timeCounter;
+    if(duration > 60)
+    {
+        cusData->timeCounter = duration_cast<seconds>(system_clock::now().time_since_epoch()).count();
+
+        double newPutFrameErrors = (double)stream_metrics_raw->putFrameErrors - cusData->totalPutFrameErrorCount;
```
disa6302, line 330: Number of errors need not be double. It can be UINT64. Only rate needs to be DOUBLE.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+            pushMetric("PutFrameErrorRate", putFrameErrorRate, Aws::CloudWatch::Model::StandardUnit::Count_Second, metricDatum, cusData->pAggregated_dimension, cwRequest);
+            pushMetric("ErrorAckRate", errorAckRate, Aws::CloudWatch::Model::StandardUnit::Count_Second, metricDatum, cusData->pAggregated_dimension, cwRequest);
+            pushMetric("TotalNumberOfErrors", totalNumberOfErrors, Aws::CloudWatch::Model::StandardUnit::Count, metricDatum, cusData->pAggregated_dimension, cwRequest);
+        }
+
+        cusData->pCanaryLogs->canaryStreamSendLogs(cusData->pCloudwatchLogsObject);
+    }
+
+    // Send metrics to CW
+    auto outcome = cusData->pCWclient->PutMetricData(cwRequest);
```
disa6302, line 357: Replace with Async call.

disa6302: Also, what metric is being pushed here? Shouldnt this be done for every stream metric and other metrics?

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+            sessionTokenStr = "";
+        }
+
+        data->credential.reset(new Credentials(string(accessKey),
+                                               string(secretKey),
+                                               sessionTokenStr,
+                                               std::chrono::seconds(DEFAULT_CREDENTIAL_EXPIRATION_SECONDS)));
+        credential_provider.reset(new CanaryCredentialProvider(*data->credential.get()));
+
+    } else if (nullptr != (iot_get_credential_endpoint = getenv("IOT_GET_CREDENTIAL_ENDPOINT")) &&
```
disa6302, line 591: Use `GETENV` instead of `getenv`

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+    unique_ptr<CredentialProvider> credential_provider;
+
+    if (nullptr == (defaultRegion = getenv(DEFAULT_REGION_ENV_VAR))) {
+        defaultRegionStr = DEFAULT_AWS_REGION;
+    } else {
+        defaultRegionStr = string(defaultRegion);
+    }
+    LOG_INFO("Using region: " << defaultRegionStr);
+
+    if (nullptr != (accessKey = getenv(ACCESS_KEY_ENV_VAR)) &&
```
disa6302, line 573: Think if can push this credential configuration within CanaryConfig.cpp. The purpose CanaryConfig.cpp should be to do this kind of env based config set up. Both static and IoT credential provider. See if it can be done as part of `initConfigWithEnvVars`

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+    // reset state
+    data->stream_status = STATUS_SUCCESS;
+    data->stream_started = false;
+
+
+    LOG_DEBUG("Stream is ready");
+}
+
+// callback when each RTSP stream has been created
+static void pad_added_cb(GstElement *element, GstPad *pad, GstElement *target) {
```
disa6302, line 673: Not needed at the moment. We do not support rtsp yet in our canaries.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+
+    // videotestsrc must be set to "live" in order for pts and dts to be incremented
+    g_object_set(source, "is-live", TRUE, NULL);
+
+    // configure appsink
+    g_object_set(G_OBJECT (appsink), "emit-signals", TRUE, "sync", FALSE, NULL);
+    g_signal_connect(appsink, "new-sample", G_CALLBACK(on_new_sample), data);
+    
+    // define the elements
+    // h264enc = gst_element_factory_make("vtenc_h264_hw", "h264enc");
```
disa6302, line 715: Do we not need this? If not, remove.

canary/producer-cpp/src/Include.h
```diff
@@ -0,0 +1,71 @@
...
+// #define DEFAULT_FRAGMENT_DURATION_MILLISECONDS 2000
+#define DEFAULT_TIMECODE_SCALE_MILLISECONDS 1
+#define DEFAULT_KEY_FRAME_FRAGMENTATION TRUE
+#define DEFAULT_FRAME_TIMECODES TRUE
+#define DEFAULT_ABSOLUTE_FRAGMENT_TIMES TRUE
+#define DEFAULT_FRAGMENT_ACKS TRUE
+#define DEFAULT_RESTART_ON_ERROR TRUE
+#define DEFAULT_RECALCULATE_METRICS TRUE
+#define DEFAULT_AVG_BANDWIDTH_BPS (4 * 1024 * 1024)
+// #define DEFAULT_BUFFER_DURATION_SECONDS 120
```
disa6302, line 62: Remove if not required. Also, If we plan to have an Include.h, I would suggest moving #defines from all the files into this and have all header files reference Include.h file. It is more modular and cleaner.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,899 @@
...
+
+    GstElement *pipeline;
+    int ret;
+    GstStateChangeReturn gst_ret;
+
+    // Reset first frame pts
+    data->first_pts = GST_CLOCK_TIME_NONE;
+
+    switch (data->streamSource) {
+        case CustomData::TEST_SOURCE:
```
disa6302, line 769: Would suggest changing the place for definition of TEST_SOURCE outside CustomData to avoid this.

canary/producer-cpp/src/CanaryConfig.cpp
```diff
@@ -0,0 +1,75 @@
...
+    bufferDuration = DEFAULT_BUFFER_DURATION_SECONDS;
+    storageSizeInBytes = 0;
+    useAggMetrics = true;
+}
+
+void CanaryConfig::setEnvVarsString(string &configVar, string envVar)
+{
+    if (getenv(envVar.c_str()) != NULL)
+    {
+        configVar = getenv(envVar.c_str());
```
disa6302, line 23: Use PIC functions. GETENV instead of getenv.

canary/producer-cpp/src/CanaryConfig.cpp
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+void CanaryConfig::setEnvVarsInt(int &configVar, string envVar)
+{
+    if (getenv(envVar.c_str()) != NULL)
+    {
+        configVar = stoi(getenv(envVar.c_str()));
```
disa6302, line 31: For string to int or other data type conversions, reference this: https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/utils/include/com/amazonaws/kinesis/video/utils/Include.h. There is a strtoint() function that does this.

canary/producer-cpp/src/CanaryLogs.cpp
```diff
@@ -0,0 +1,113 @@
...
+    auto outcome = pCloudwatchLogsObject->pCwl->PutLogEvents(request);
+    if (!outcome.IsSuccess()) {
+        DLOGE("Failed to push logs: %s", outcome.GetError().GetMessage().c_str());
+    } else {
+        DLOGS("Successfully pushed logs to cloudwatch");
+    }
+    pCloudwatchLogsObject->canaryInputLogEventVec.clear();
+}
+
+// VOID CanaryLogs::cloudWatchLogger(UINT32 level, PCHAR tag, PCHAR fmt, ...)
```
disa6302, line 93: Remove if not needed?

canary/producer-cpp/src/Canary.cpp
stefankiesz: Isn't this a reference? https://www.ibm.com/docs/en/zos/2.4.0?topic=calls-pass-by-reference-c-only

canary/producer-cpp/src/Canary.cpp
stefankiesz: All the metrics are added to the cwRequest to be sent all together in the cwRequest, and it is done in four places in the code: upon persisted ack, upon received ack, upon keyframe, and for startup latency.

stefankiesz: (In the case above, it is for all the metrics upon keyframe)

stefankiesz: And for AmazonCloudWatchAsync calls I am getting: 
`error: no matching function for call to ‘Aws::CloudWatch::CloudWatchClient::PutMetricDataAsync(Aws::CloudWatch::Model::PutMetricDataRequest&)`
Is there a different namespace I should be using that includes this? Everything I'm seeing online on this is for Java.

canary/producer-cpp/src/Canary.cpp
stefankiesz: I moved all of the GETENV calls to CanaryConfig.cpp which cleaned it up some, but the rest of the function must stay outside of the CanaryConfig class because it involves CustomData members, and CanaryConfig does not have ability to access CustomData (only vis versa), I can make them point to each other if that would be better.

canary/producer-cpp/src/Canary.cpp
stefankiesz: Separated client from stream, and I also modularized error metrics push into its own function too.

canary/producer-cpp/src/Include.h
stefankiesz: Having Include.h file in all the other headers would not work because they are all being aggregated to be included into the Include.h file - this aggregation is necessary because there are classes that refer to other classes.

canary/producer-cpp/src/Canary.cpp
stefankiesz: Fixed this for everything but the iot credential and canary callback stuff - seems to make more sense to leave as is for these two.

canary/producer-cpp/jobs/orchestrator.groovy
stefankiesz: Not needed, but it seemed like a useful thing to keep from the C Canary, it makes it easy to set the typical durations, delays, etc. I set the env vars for Canary duration this using these definitions

canary/producer-cpp/jobs/orchestrator.groovy
```diff
@@ -0,0 +1,214 @@
...
+
+                            def gitHash = sh(returnStdout: true, script: 'git rev-parse HEAD')
+                            COMMON_PARAMS << string(name: 'GIT_HASH', value: gitHash)
+                        }
+
+                        // TODO: Use matrix to provide configurability in parameters
+                        build(
+                            job: NEXT_AVAILABLE_RUNNER,
+                            parameters: COMMON_PARAMS + [
+                                string(name: 'CANARY_STREAM_NAME', value: "Continuous-Longrun-01"),
```
disa6302, line 118: As epr our discussion, can we reduce the number of builds to unique runs? As per this, we would only need a Realtime long run and realtime periodic. Also change the way the stream names are configured. Remove the `-01`, `-02`, etc.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,864 @@
...
+}  // namespace video
+}  // namespace kinesis
+}  // namespace amazonaws
+}  // namespace com;
+
+
+// add frame pts, frame index, original frame size, CRC to beginning of buffer
+VOID addCanaryMetadataToFrameData(PFrame pFrame)
+{
+    PBYTE pCurPtr = pFrame->frameData;
```
disa6302, line 231: Doing this would overwrite the gstreamer pFrame->frameData. You would have to memcopy the GStreamer frame data to the location starting from CANARY_METADATA_SIZE to ensure the header is appended at the start.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,864 @@
...
+    lastKeyFrameTime = curKeyFrameTime;
+}
+
+VOID pushErrorMetrics(Frame frame, CustomData *cusData, double duration)
+{
+    Aws::CloudWatch::Model::MetricDatum metricDatum;
+    Aws::CloudWatch::Model::PutMetricDataRequest cwRequest;
+    cwRequest.SetNamespace("KinesisVideoSDKCanary");   
+
+    //auto stream_metrics = cusData->kinesisVideoStream->getMetrics();
```
disa6302, line 278: Remove.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,864 @@
...
+    // define the elements
+    source = gst_element_factory_make("videotestsrc", "source");
+    autovidcon = gst_element_factory_make("autovideoconvert", "vidconv");
+    h264enc = gst_element_factory_make("x264enc", "h264enc");
+    h264parse = gst_element_factory_make("h264parse", "h264parse");
+    appsink = gst_element_factory_make("appsink", "appsink");
+    h264parse = gst_element_factory_make("h264parse", "h264parse");
+
+    // to change output video pattern to a moving ball, uncomment below
+    //g_object_set(source, "pattern", 18, NULL);
```
disa6302, line 674: Remove / uncomment.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,864 @@
...
+
+    if (cusData->pCanaryConfig->useAggMetrics)
+    {
+        pushMetric("PutFrameErrorRate", putFrameErrorRate, Aws::CloudWatch::Model::StandardUnit::Count_Second, metricDatum, cusData->pAggregatedDimension, cwRequest);
+        pushMetric("ErrorAckRate", errorAckRate, Aws::CloudWatch::Model::StandardUnit::Count_Second, metricDatum, cusData->pAggregatedDimension, cwRequest);
+        pushMetric("TotalNumberOfErrors", totalNumberOfErrors, Aws::CloudWatch::Model::StandardUnit::Count, metricDatum, cusData->pAggregatedDimension, cwRequest);
+    }
+
+    // Send metrics to CW
+    cusData->pCWclient->PutMetricDataAsync(cwRequest, onPutMetricDataResponseReceivedHandler);
```
disa6302, line 305: Again. Not sure I understand this. Using pushMetric, you are setting cwRequest.AddMetricDatum = metricDatum. This call to `pushMetric` happens for every unique metric, however, your PutMetricDataAsync call is made once? Wouldnt the cwRequest get overwritten? With this call, it seems like you are actually only publishing only aggregated metrics if useAggMetrics is set and totalNumberOfErrors if not. You will need this line after every metric.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,864 @@
...
+
+                auto currentTimestamp = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
+                auto receivedAckLatency = (currentTimestamp - timeOfFragmentEndSent); // [milliseconds]
+                pushMetric("ReceivedAckLatency", receivedAckLatency, Aws::CloudWatch::Model::StandardUnit::Milliseconds, receivedAckLatencyDatum, data->pDimensionPerStream, cwRequest);
+                LOG_DEBUG("Received Ack Latency: " << receivedAckLatency);
+                if (data->pCanaryConfig->useAggMetrics)
+                {
+                    pushMetric("ReceivedAckLatency", receivedAckLatency, Aws::CloudWatch::Model::StandardUnit::Milliseconds, receivedAckLatencyDatum, data->pAggregatedDimension, cwRequest);
+                }
+                data->pCWclient->PutMetricDataAsync(cwRequest, onPutMetricDataResponseReceivedHandler);
```
disa6302, line 205: Same about putMetricDataAsync. Please fix this everywhere. We need per stream metrics as well and not just aggregated. Include the call as part of `pushMetric` API so that the `PutMetricDataAsync` is invoked only at one location. Makes code cleaner.

canary/producer-cpp/jobs/canary_seed.groovy
```diff
@@ -0,0 +1,84 @@
...
+import jenkins.model.*;
+import org.jenkinsci.plugins.scriptsecurity.scripts.*;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.*;
+
+WORKSPACE="."
+GROOVY_SCRIPT_DIR="$WORKSPACE/canary/producer-cpp/jobs"
+DAYS_TO_KEEP_LOGS=2
+NUMBER_OF_LOGS=1
+MAX_EXECUTION_PER_NODE=1
+NAMESPACE="producer"
```
disa6302, line 11: Namespace should be changed to producercpp

canary/producer-cpp/src/CanaryLogs.cpp
```diff
@@ -0,0 +1,91 @@
...
+    auto request = Aws::CloudWatchLogs::Model::PutLogEventsRequest()
+                       .WithLogGroupName(pCloudwatchLogsObject->logGroupName)
+                       .WithLogStreamName(pCloudwatchLogsObject->logStreamName)
+                       .WithLogEvents(pCloudwatchLogsObject->canaryInputLogEventVec);
+    if (pCloudwatchLogsObject->token != "") {
+        request.SetSequenceToken(pCloudwatchLogsObject->token);
+    }
+    auto outcome = pCloudwatchLogsObject->pCwl->PutLogEvents(request);
+    if (!outcome.IsSuccess()) {
+        DLOGE("Failed to push logs: %s", outcome.GetError().GetMessage().c_str());
```
disa6302, line 86: Change from DLOGx to LOG_* .

canary/producer-cpp/jobs/canary_seed.groovy
stefankiesz: Just temporarily changed it back to producer for testing something

canary/producer-cpp/src/Canary.cpp
```diff
@@ -0,0 +1,865 @@
...
+
+}  // namespace video
+}  // namespace kinesis
+}  // namespace amazonaws
+}  // namespace com;
+
+// add frame pts, frame index, original frame size, CRC to beginning of buffer
+VOID addCanaryMetadataToFrameData(PFrame pFrame)
+{
+    PBYTE pBuf = new BYTE[CANARY_METADATA_SIZE + pFrame->size];
```
disa6302, line 232: You can avoid this by just memcpy'ing in `create_kinesis_video_frame`. frameData is being reinterpreted there. All you need to do is `MEMCPY(frame->frameData + METADATA_SIZE, frame->frameData, len) and set `frame->size` to include `METADATA_SIZE` and it will be handled. Then invoking this function is as it is done here. Your implementation would lead to memory leak because you do not clear the memory.

disa6302 (changes requested):
Good job. Let us just fix the frame data populating portion. Also, please remove usage of couts and use the logger instead as added in previous comments :)

disa6302 (approved):
Yay!

PR #275, opened 2023-03-06
Kvssink canary cpp demos latency

NikunjSarda:
*Issue #, if available:*

N/A

*Description of changes:*

Added startup latency metric, IOT credential support for Jenkins.

Changes:

1. Startup latency metric is now pushed to cloudwatch, the metric being calculated on first successful frame put through kvssink, a gstreamer signal is emitted in kvssink for the first successful frame which is connected in demos canary using gstreamer connect

2. IOT credential support for Jenkins is added, the IOT credential is working in demos as per earlier changes made while migrating the demos canary to kvssink, now the same thing could be used while using Jenkins server, for the above implementation the script changes are done in Job scripts, a certificate setup script is run as part of Job script to setup environment variable for each run for IOT, added IOT elements(like endpoint, certificate, ...) to retrieve environment setup in earlier step for IOT use
 
*Testing:*

Tested with both AWS credentials and IOT credentials in local and Jenkins, the startup latency metric and other metrics are visible in the AWS account cloudwatch, the IOT credential for Jenkins also publish cloudwatch metrics and the default video media play in AWS account Kinesis Video Stream for the stream created, tested using local(MAC) as Jenkins server and node server set on EC2 instance(Ubuntu 18.04)


canary/producer-cpp/src/Canary.cpp
```diff
@@ -666,15 +351,24 @@ int gstreamer_test_source_init(CustomData *data, GstElement *pipeline) {
...
-    // configure appsink
-    g_object_set(G_OBJECT (appsink), "emit-signals", TRUE, "sync", FALSE, NULL);
-    g_signal_connect(appsink, "new-sample", G_CALLBACK(on_new_sample), data);
+    // configure kvssink
+    g_object_set(G_OBJECT (kvssink), "stream-name", data->streamName, "storage-size", 128, NULL);
+
+    string use_iot_cred(data->pCanaryConfig->use_iot_credential_provider);
+    transform(use_iot_cred.begin(), use_iot_cred.end(),use_iot_cred.begin(), ::tolower);
+
+    if(use_iot_cred.compare("true") == 0){
```
disa6302, line 366: This block seems weird. determine should be invoked either ways. Basically, the check for iot or not should happen within `determine_credentials`. If iot and if the required isnt set, the job should fail. If it is not IoT, access and secret should be set and if it isnt set, we need to fail. In either cases, we should push a status code to the cloudwatch (preferred) or log and fail the job.

canary/producer-cpp/src/Canary.cpp
```diff
@@ -666,15 +351,24 @@ int gstreamer_test_source_init(CustomData *data, GstElement *pipeline) {
...
+        determine_credentials(kvssink, data);
+    }
+
+    g_signal_connect(G_OBJECT(kvssink), "stream-client-metric", (GCallback) start_put_frame_kvs, data);
```
disa6302, line 370: Camel case throughout and use the suffix handler since this is a callback.

disa6302 (changes requested):

PR #296, opened 2023-07-19
WebRTC Master-End Metrics

stefankiesz:
The Canary is currently using local sample frames so they are viewable from KVS stream playback console for testing. The following env vars were set for testing:
```
export CANARY_CHANNEL_NAME=<storageChannelName>
export CANARY_IS_MASTER=TRUE
export CANARY_TRICKLE_ICE=TRUE
```
> **Note**
> The storageChannelName channel must be configured to ingest media into a KVS stream using [these](https://docs.aws.amazon.com/kinesisvideostreams-webrtc-dg/latest/devguide/configure-ingestion.html) instructions. 

The Canary has not yet been tested on a Jenkins server. Proper aggregate metric labeling can be tested once run on a Jenkins server.


stefankiesz:
Switching to a different branch for development

PR #304, opened 2023-08-01
Use multi-stage build to reduce image size

yuma-m:
*Issue #, if available:* N/A

*Description of changes:* Use [multi-stage build](https://docs.docker.com/build/building/multi-stage/) to reduce the image size. Reducing image size from 1.2GiB to 653MiB.

```
REPOSITORY        TAG       IMAGE ID        CREATED           PLATFORM       SIZE         BLOB SIZE
rtspdockertest    latest    d34fbf828bfb    24 hours ago      linux/arm64    1.2 GiB      494.1 MiB   // original
rtsp-slim         latest    b82b0ef5ba7b    10 minutes ago    linux/arm64    653.1 MiB    292.4 MiB   // reduced
```

### DIFF
--- a/producer-cpp/docker-rtsp/Dockerfile
+++ b/producer-cpp/docker-rtsp/Dockerfile
@@ -9,18 +9,12 @@ RUN apt-get upgrade && \
     g++ \
     git \
     gstreamer1.0-plugins-base-apps \
-    gstreamer1.0-plugins-bad \
-    gstreamer1.0-plugins-good \
-    gstreamer1.0-plugins-ugly \
     libgstreamer1.0-dev \
     libgstreamer-plugins-base1.0-dev \
     m4 \
     maven \
     openjdk-8-jdk \
-    python2.7 \
     pkg-config \
-    vim \
-    wget  \
     xz-utils && \
     rm -rf /var/lib/apt/lists/* && \
     cd /opt/
@@ -29,7 +23,7 @@ RUN apt-get upgrade && \
 WORKDIR /opt/
 RUN git clone https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp.git
 WORKDIR /opt/amazon-kinesis-video-streams-producer-sdk-cpp/build/
-RUN cmake .. -DBUILD_GSTREAMER_PLUGIN=ON &&\
+RUN cmake .. -DBUILD_GSTREAMER_PLUGIN=ON && \
     make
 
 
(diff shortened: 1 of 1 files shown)


producer-cpp/docker-rtsp/Dockerfile
```diff
@@ -32,6 +32,22 @@ WORKDIR /opt/amazon-kinesis-video-streams-producer-sdk-cpp/build/
...
+
+FROM ubuntu:18.04
+
+RUN apt update && \
+    apt install -y \
+    gstreamer1.0-plugins-base-apps \
+    gstreamer1.0-plugins-bad \
+    gstreamer1.0-plugins-good \
+    gstreamer1.0-plugins-ugly \
+    openjdk-8-jre &&\
```
disa6302, line 44: Nit: Space between && and \

producer-cpp/docker-rtsp/Dockerfile
```diff
@@ -32,6 +32,22 @@ WORKDIR /opt/amazon-kinesis-video-streams-producer-sdk-cpp/build/
 RUN cmake .. -DBUILD_GSTREAMER_PLUGIN=ON &&\
```
disa6302, line 32: Although there is no change to this line as part of this PR, can we include a space between && and \

producer-cpp/docker-rtsp/Dockerfile
disa6302: Also, curious. Is there a reason we are installing gstreamer again? Is it not possible to use `FROM builder` instead to avoid dual installations?

disa6302 (commented):
This is great! Thanks @yuma-m ! Have a few comments.

producer-cpp/docker-rtsp/Dockerfile
yuma-m: Yes, it is possible to copy GStreamer related files from builder. I think some of these packages are not required for building SDK, I will check and optimize these duplications.

yuma-m:
@disa6302 I've fixed where you commented and removed packages not required for building SDK.

sirknightj (approved):

producer-cpp/docker-rtsp/Dockerfile
```diff
@@ -29,9 +23,24 @@ RUN apt-get upgrade && \
...
 RUN git clone https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp.git
 WORKDIR /opt/amazon-kinesis-video-streams-producer-sdk-cpp/build/
-RUN cmake .. -DBUILD_GSTREAMER_PLUGIN=ON &&\
+RUN cmake .. -DBUILD_GSTREAMER_PLUGIN=ON && \
     make
 
+
+FROM ubuntu:18.04
+
+RUN apt update && \
```
disa6302, line 32: Quick question. 
We would need the gstreamer install steps in the BUILDER right and not here?
Wouldnt the build `cmake .. -DBUILD_GSTREAMER_PLUGIN=ON && \` fail because gstreamer is not installed? Have we tried building it after the new changes?

disa6302 (approved):
Thanks Yuma! Merging the PR.

PR #308, opened 2023-08-10
Profiling canary

disa6302:
Issue #, if available: N/A

Description of changes:

What was changed?

Added a new periodic job run for WebRTC that is solely dedicated to getting metrics contributing to start up latency.

Why was it changed?

To allow for easy visualization of changing start up latency metrics over time

How it was changed?

Modified the app to pull the metrics on a thread. The metrics will be extracted and pushed to cloudwatch after the first frame is sent out. A new thread is used to ensure all the metrics are pushed out. It was initially added in callbacks which led to missing metrics because cloudwatch was slow to process it.

Testing:

- Local testing on jenkins server for a few days


hassanctech (approved):
LGTM!

PR #313, opened 2023-08-26
Upgrade AWS CPP SDK Version for WebRTC-C Canary

stefankiesz:
## Changes
Updates AWS CPP SDK version from **1.9.67** to **1.11.143** in WebRTC-C Canary.
- `CMAKE_PREFIX_PATH` is now set, and it is set to the dependency build directory `/build/_deps`
  - This solved `undefined reference to 'EVP_idea_cbc@OPENSSL_1_1_0'` errors associated with the following warning by directly specifying which directory to search in for dependencies.
```
Cannot generate a safe runtime search path for target [...] because files in some directories
may conflict with libraries in implicit directories:
    runtime library [libcrypto.so.1.1] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/ubuntu/amazon-kinesis-video-streams-demos/canary/webrtc-c/build/_deps/webrtc-src/open-source/lib
Some of these libraries may not be found correctly.  
```
- `/build/_deps/lib` is then added to `link_directories` to point to the dependency libraries, preventing the following errors:
```
/usr/bin/ld: cannot find -lkvspicUtils
/usr/bin/ld: cannot find -lkvsCommonLws
/usr/bin/ld: cannot find -lwebsockets
```

## Testing
These changes have been tested by successfully running `cmake`, `make`, and `./kvsWebrtcCanaryWebrtc` and then verifying that metrics are emitting to CloudWatch.

<br>
<br>

### DIFF
--- a/canary/webrtc-c/CMakeLists.txt
+++ b/canary/webrtc-c/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.13)
 project(KVSWebRTCCanary LANGUAGES C CXX)
 
 set(CMAKE_CXX_STANDARD 11)
@@ -42,9 +42,13 @@ add_definitions(-DCMAKE_DETECTED_CACERT_PATH)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-core/include)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-monitoring/include)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-logs/include)
+
 include_directories(${webrtc_SOURCE_DIR}/src/include)
 include_directories(${webrtc_SOURCE_DIR}/open-source/include)
 link_directories(${webrtc_SOURCE_DIR}/open-source/lib)
+
+link_directories(${CMAKE_CURRENT_SOURCE_DIR}/build/_deps/lib)
+
 add_library(
   kvsWebrtcCanary
   src/Config.cpp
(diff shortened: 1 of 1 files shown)


canary/webrtc-c/CMakeLists.txt
```diff
@@ -42,9 +50,13 @@ add_definitions(-DCMAKE_DETECTED_CACERT_PATH)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-core/include)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-monitoring/include)
 include_directories(${cloudwatch_SOURCE_DIR}/aws-cpp-sdk-logs/include)
+
 include_directories(${webrtc_SOURCE_DIR}/src/include)
 include_directories(${webrtc_SOURCE_DIR}/open-source/include)
 link_directories(${webrtc_SOURCE_DIR}/open-source/lib)
+
+link_directories(${CMAKE_CURRENT_SOURCE_DIR}/build/_deps/lib)
```
hassanctech, line 58: It looks like this path is `OPEN_SRC_INSTALL_PREFIX` as defined in LN15 above. Let's use that variable directly so we don't have to change it in two places in case we decided to change the install path.

stefankiesz: Sounds good, testing build now to ensure it still works 👍

disa6302 (approved):

sirknightj (approved):

canary/webrtc-c/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.13)
```
sirknightj, line 1: Thoughts on bumping this all the way to most recent?

PR #316, opened 2023-09-04
Update WebRTC Storage Canary

stefankiesz:
## Overview
WebRTC Storage Canary Updates:
- Revamps consumer-end Canary
  - Improves code modularity
  - Adds `TimeToFirstFragment` metric
- Updates webrtc-end Canary to align with upcoming WebRTC-C Storage SDK release
  - `JoinStorageSession` no longer called within peer-side app
  - Removes use of any customer-facing MediaServer-related API
## Consumer-End Changes
- Separates the two consumer-end metric calculations `FragmentContinuity` and `TimeToFirstFragment` into individual methods which are called based on the configured CANARY_LABEL
- Moves CloudWatch metric publishing from within the calculation methods to a designated method
- Fixes Consumer Canary issue of relying on `System.exit(0)` to exit the application
  - Discovered that active thread count was not being decremented after calling `PutMetricDataRequest` to CloudWatch, perhaps due to a thread completion indicator not getting emitted after the Timer Task increments to next iteration
  - CloudWatch client is switched from async version to synchronous to solve this
## WebRTC-End Changes
- Removes all changes made to `Peer` class
- Removes `useMediaStorage` flag from Canary as it is not longer used by `Peer`
- Introduces `isStorage` configurable parameter to replace `useMediaStorage` for CloudWatch metric-labeling purposes
- Removes the termination of Canary upon disconnection from peer to allow for automatic reconnect after the 60min connection limit
### Consumer-End
Set env vars:
```
export CANARY_STREAM_NAME=<storage-configured streamName>
export CANARY_DURATION_IN_SECONDS=<300 or more for Media Server compatibility>
export CANARY_LABEL=<WebrtcLongRunning or WebrtcPeriodic>
export AWS_DEFAULT_REGION=us-west-2
```
Run:
```
java -classpath target/aws-kinesisvideo-producer-sdk-canary-consumer-1.0-SNAPSHOT.jar:$(cat tmp_jar) -Daws.accessKeyId=${AWS_ACCESS_KEY_ID} -Daws.secretKey=${AWS_SECRET_ACCESS_KEY} com.amazon.kinesis.video.canary.consumer.WebrtcStorageCanaryConsumer
```

### WebRTC-End
Set env vars:
```
export CANARY_CHANNEL_NAME=<storage-configured channelName>
export CANARY_IS_MASTER=TRUE
export CANARY_TRICKLE_ICE=TRUE
export AWS_KVS_LOG_LEVEL=1
[...]

PR #328, opened 2023-11-17
Sample-Based WebRTC Storage Canary

stefankiesz:
## Overview
Introduces the Media Server WebRTC with Storage Canary. The canary consists of two canary apps: a master end and a consumer end. The master canary is a modified version of the **kvsWebRTCClientMaster** sample which includes the following changes:
- Change C files to C++ to use relevant canary files and libraries
- Integrate with existing canary CloudWatch architecture
- Add configurability using environment variables
- Add and manage timers and threads for metrics publishing and duration scheduling
- Use CloudWatch logging rather than file logger

For a diff of the master sample changes, refer to [this](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1847) draft PR.

The WebRTC canary Jenkins runner job was restructured to build both master and consumer ends in a modular fashion. Some changes to the canary described in [this](https://github.com/aws-samples/amazon-kinesis-video-streams-demos/pull/316) unmerged PR are also included.

The following metrics are emitted by the canary:
- RTP outbound metrics 
- Profiling metrics
- Exit status code
- E2E latency between initiating master end and receiving first frame in consumer end (periodic case only)
- E2E latency between master end sending first frame to Media Server and receiving first frame in consumer end (periodic case only)
- Time taken to send out first frame after a 60min Media Server disconnect
- Ingested fragment continuity (long-run case only)

<br>

### WebRTC-End
Set env vars:
```
export CANARY_CHANNEL_NAME=<storage-configured channelName>
export CANARY_IS_MASTER=TRUE
export CANARY_TRICKLE_ICE=TRUE
export AWS_KVS_LOG_LEVEL=2
export CANARY_DURATION_IN_SECONDS=<300 or more for Media Server compatibility>
export CANARY_LABEL=<WebrtcLongRunning or WebrtcPeriodic>
```

Run:
```
./kvsWebrtcStorageSample
```

### Consumer-End
Set env vars:
```
export CANARY_STREAM_NAME=<storage-configured streamName>
export CANARY_DURATION_IN_SECONDS=<300 or more for Media Server compatibility>
export CANARY_LABEL=<WebrtcLongRunning or WebrtcPeriodic>
export AWS_DEFAULT_REGION=us-west-2
```
Run:
```
java -classpath target/aws-kinesisvideo-producer-sdk-canary-consumer-1.0-SNAPSHOT.jar:$(cat tmp_jar) -Daws.accessKeyId=${AWS_ACCESS_KEY_ID} -Daws.secretKey=${AWS_SECRET_ACCESS_KEY} com.amazon.kinesis.video.canary.consumer.WebrtcStorageCanaryConsumer
```

<br>
<br>


stefankiesz:
I'm leaving extra logs and print statements in for now until after testing any requested changes.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/RealTimeFrameProcessor.java
```diff
@@ -0,0 +1,86 @@
...
+        return new RealTimeFrameProcessor();
+    }
+
+    private RealTimeFrameProcessor() {
+    }
+
+    @Override
+    public void process(Frame frame, MkvTrackMetadata trackMetadata, Optional<FragmentMetadata> fragmentMetadata, Optional<FragmentMetadataVisitor.MkvTagProcessor> tagProcessor) throws FrameProcessException {
+        System.out.println("RealTimeFrameProcessor invoked for frame" + frame);
+        if (!isFirstFrame) {
```
stefankiesz, line 48: Should add a check to only process the metrics if the stream has been empty, otherwise the consumer end may have started too late to catch the true "first frame received" time

stefankiesz:
TODO:
- Decide on run interval lengths and nomenclature
- Decide on consumer metric interval lengths
- Make consumer e2e metrics safer (check for late consumer end starts, etc.)
- Finalize what metrics are to be tracked in the dashboard and fix a typo in the description for `Offer Received to First Frame Sent (TimeToFirstFrame)`

stefankiesz:
The webrtc c sdk PR draft for diff comparison: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1847

canary/webrtc-c/src/media-server-storage/Common.cpp
```diff
@@ -0,0 +1,1864 @@
...
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PSampleConfiguration pSampleConfiguration;
+    UINT32 i;
+    UINT64 data;
+    StackQueueIterator iterator;
+    BOOL locked = FALSE;
+
+    std::string filePath = "../toConsumer.txt";
```
disa6302, line 1328: This file path should be set up as an env. We should use envs in places where both webrtc and consumer depend on the value to be the same so that we do not need to update the C files and Java files in the future which could cause sync issues in case of a miss in modifying required code.

canary/webrtc-c/src/media-server-storage/Common.cpp
```diff
@@ -0,0 +1,1864 @@
...
+
+    if (IS_VALID_MUTEX_VALUE(pSampleConfiguration->signalingSendMessageLock)) {
+        MUTEX_FREE(pSampleConfiguration->signalingSendMessageLock);
+    }
+
+    if (IS_VALID_CVAR_VALUE(pSampleConfiguration->cvar)) {
+        CVAR_FREE(pSampleConfiguration->cvar);
+    }
+
+#ifdef IOT_CORE_ENABLE_CREDENTIALS
```
disa6302, line 1418: We are setting this up with env for our normal canaries, correct? We are not testing the IoT path for media server I assume?

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
```diff
@@ -0,0 +1,367 @@
+#include "Samples.h"
+#include "../Include.h"
+
+
+extern PSampleConfiguration gSampleConfiguration;
+
+VOID shceduleShutdown(UINT64 duration, PSampleConfiguration pSampleConfiguration)
```
disa6302, line 7: Nit: `scheduleShutdown`

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
```diff
@@ -0,0 +1,367 @@
...
+    DLOGI("[KVS Master] KVS WebRTC initialization completed successfully");
+    initialized = TRUE;
+
+    CHK_STATUS(initSignaling(pSampleConfiguration, SAMPLE_MASTER_CLIENT_ID));
+    DLOGI("[KVS Master] Channel %s set up done ", pChannelName);
+
+    if(canaryConfig.duration.value != 0) {
+        DLOGD("[Canary] Scheduling canary duration for %lu seconds", canaryConfig.duration.value / HUNDREDS_OF_NANOS_IN_A_SECOND);
+        canaryDurationThread = std::thread(shceduleShutdown, canaryConfig.duration.value, pSampleConfiguration);
+        canaryDurationThread.detach();
```
disa6302, line 100: No need to detach. We should join().

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
```diff
@@ -0,0 +1,367 @@
...
+    if (pSize != NULL) {
+        *pSize = (UINT32) size;
+    }
+
+    return retStatus;
+}
+
+// Save first-frame-sent time to file for consumer-end access.
+PVOID writeFirstFrameSentTimeToFile(){
+    DLOGI("[Canary] Opening toConsumer.txt");
```
disa6302, line 176: Let us use PIC APIs here given we use PIC APIs for read.

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
```diff
@@ -0,0 +1,367 @@
...
+    FILE *toConsumer = FOPEN("../toConsumer.txt", "w");
+    if (toConsumer == NULL)
+    {
+        printf("[Canary] Error opening toConsumer.txt\n");
+        exit(1);
```
disa6302, line 181: Not sure if we want to use exit(0) here. Also, why is this function a PVOID()? Dont see this being run in a thread.

disa6302 (changes requested):
General PR comments:

1. Check for hardcoded values and ensure these are set up in a separate `Constants.java` file or `#define` in case of C. 
2. I see a lot of usage of System.exit(0). Let us ensure we do not use this unless there is an absolute necessity to do so. We should gracefully handle situations wherever applicable. 
3. Let us reevaluate the need to use static methods, static members, protected members and confirm it is all indeed necessary.
4. Handle shutdown and cleanup of AWS resources wherever applicable. Revisit the class documents for the AWS resources being used.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
```diff
@@ -0,0 +1,107 @@
...
+    https://github.com/aws/amazon-kinesis-video-streams-parser-library/blob/0056da083402a96cffcdfd8bca9433dd25eb7470/src/main/java/com/amazonaws/kinesisvideo/parser/examples/ListFragmentWorker.java#L7
+
+    This version returns Fragment objects rather than only the Fragment Number
+ */
+
+@Log4j2
+public class CanaryListFragmentWorker implements Callable {
+    private final FragmentSelector fragmentSelector;
+    private final AmazonKinesisVideoArchivedMedia amazonKinesisVideoArchivedMedia;
+    private final long fragmentsPerRequest = 100;
```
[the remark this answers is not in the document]
stefankiesz: This is a carry-over from our [ListFragmentWorker example](https://github.com/aws/amazon-kinesis-video-streams-parser-library/blob/master/src/main/java/com/amazonaws/kinesisvideo/parser/examples/ListFragmentWorker.java#L22). I don't think there is a particular reason other than making use of the pagination feature, breaking up results to get them 100 at a time might be more efficient than gathering all fragments of a large request.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
```diff
@@ -0,0 +1,107 @@
...
+    protected final String streamName;
```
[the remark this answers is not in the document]
stefankiesz: Not needed, nice catch.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
```diff
@@ -0,0 +1,107 @@
...
+            for (Fragment f : result.getFragments()) {
+                fragments.add(new CanaryFragment(f));
+            }
+            String nextToken = result.getNextToken();
+
+            /* If result is truncated, keep making requests until nextToken is empty */
+            while (nextToken != null) {
+                request = new ListFragmentsRequest()
+                        .withStreamName(streamName).withNextToken(nextToken);
+                result = amazonKinesisVideoArchivedMedia.listFragments(request);
```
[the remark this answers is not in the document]
stefankiesz: If that API throws an exception, would it not exit the loop and jump to the catch block of the try block which the while loop is in?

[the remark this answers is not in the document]
stefankiesz: Yes, just tested it by forcing the loop with an invalid stream name. The loop exits and throws the following error:
```
Failure in CanaryListFragmentWorker for streamName com.amazonaws.services.kinesisvideo.model.ResourceNotFoundException: The requested stream is not found or not active. (Service: AmazonKinesisVideoArchivedMedia; Status Code: 404; Error Code: ResourceNotFoundException; Request ID: e7664a56-79cc-449b-8a03-21a0b0d1924a)
12:50:55.038 [Thread-2] ERROR com.amazon.kinesis.video.canary.consumer.CanaryListFragmentWorker - Failure in CanaryListFragmentWorker for streamName aTestStream com.amazonaws.services.kinesisvideo.model.ResourceNotFoundException: The requested stream is not found or not active. (Service: AmazonKinesisVideoArchivedMedia; Status Code: 404; Error Code: ResourceNotFoundException; Request ID: e7664a56-79cc-449b-8a03-21a0b0d1924a)
```

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
```diff
@@ -0,0 +1,107 @@
...
+
+            for (CanaryFragment cf : fragments) {
+                log.info("Retrieved fragment number {} ", cf.getFragment().getFragmentNumber());
+            }
+        } catch (Exception e) {
+            log.error("Failure in CanaryListFragmentWorker for streamName {} {}", streamName, e.toString());
+            throw e;
+        } finally {
+                        log.info("Retrieved {} Fragments and exiting CanaryListFragmentWorker for stream {}", fragments.size(), streamName);
+            return fragments;
```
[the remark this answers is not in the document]
stefankiesz: This function has return type `List<CanaryFragment>`. What would you prefer to return if not an empty list?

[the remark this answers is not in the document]
stefankiesz: Also, there is a chance that the list is partially populated before an exception is thrown, this would allow for that partial list to be returned.

If we want to diverge so much from the parser library example this whole class is a copy of, would you like me to fix up that example too?

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/RealTimeFrameProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+                super.publishMetricToCW("TotalTimeToFirstFrameConsumed", timeToFirstFragment,
+                        StandardUnit.Milliseconds);
+            } catch (FileNotFoundException ex) {
+                // Master-end timestamp file not found, try reducing the frequency of runs by
+                // increasing CANARY_DURATION_IN_SECONDS
+            } catch (IOException ioEx) {
+                // IO Exception
+            }
+            isFirstFrame = true;
+            System.exit(0);
```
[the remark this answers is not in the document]
stefankiesz: I've explored a few options, but this seems to be the only way to stop the getMedia calls once the firstFrame has been received. GetMediaWorker does not have a stop() function like the ContinuousGetMediaWorker, and we had decided that ContinuousGetMediaWorker would not be used for this. I can try to revisit that logic and see if we can make ContinuousGetMediaWorker work. Otherwise, we would need a custom GetMediaWorker implementation.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+ * for new fragments every "fragmentDuration * 2" The fragmentDuration is set by the encoder in
+ * configuration in Media Server.
+ * 
+ * For periodic-configured jobs, this Canary will emit TimeToFirstFragment metrics by continuously
+ * checking for consumable media from the specified stream via GetMedia calls.
+ */
+
+@Log4j2
+public class WebrtcStorageCanaryConsumer {
+    protected static Date canaryStartTime;
```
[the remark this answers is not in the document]
stefankiesz: Ok I moved the definitions of some of these up from main to be able to make them `final`.

Some are `protected` so that they may be exposed to the RealTimeFrameProcessor class.

They are `static` so that they can be referenced from static contexts in main(). I think we've previously discussed how this would be necessary if we want to avoid extrapolating WebrtcStorageCanaryConsumer to yet another new, separate class, and if they were to be declared in main(), the static functions here would need very long lists of parameters.

I guess making one more new class to solve all this is no big deal, but I wanted to avoid cluttering the existing Canary with new files which are only relevant to the storage Canary. How should we move forward?

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+    private static String canaryLabel;
+    private static String region;
+    private static EnvironmentVariableCredentialsProvider credentialsProvider;
+    private static AmazonKinesisVideo amazonKinesisVideo;
+    private static AmazonCloudWatch cwClient;
+
+    private static void calculateFragmentContinuityMetric(CanaryFragmentList fragmentList) {
+        try {
+            final GetDataEndpointRequest dataEndpointRequest = new GetDataEndpointRequest()
+                    .withAPIName(APIName.LIST_FRAGMENTS).withStreamName(streamName);
```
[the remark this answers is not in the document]
stefankiesz: If they are not static, the following build error is emitted: `non-static method calculateFragmentContinuityMetric() cannot be referenced from a static context`. I think they need to be static because although main() is a part of the WebrtcStorageCanaryConsumer class, main() is a static function, and there is no instance instantiated of the WebrtcStorageCanaryConsumer class.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+                System.exit(0);
+                break;
+            }
+            default: {
+                log.error("Env var CANARY_LABEL: {} must be set to either WebrtcLongRunning or WebrtcPeriodic",
+                        canaryLabel);
+                throw new Exception("CANARY_LABEL must be set to either WebrtcLongRunning or WebrtcPeriodic");
+            }
+        }
+    }
```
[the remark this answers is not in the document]
stefankiesz: `credentialsProvider` does not have any shutdown/free functions. `amazonKinesisVideo`'s and `cwClient`'s shutdown() descriptions:
> Shuts down this client object, releasing any resources that might be held open. This is an optional method, and callers are not expected to call it, but can if they want to explicitly release any open resources.

Let me know if you'd still like for me to add it.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+                intervalMetricsTimer.scheduleAtFixedRate(intervalMetricsTask, intervalInitialDelay, intervalDelay);
+                Thread.sleep(canaryRunTime * 1000);
+                intervalMetricsTimer.cancel();
+                break;
+            }
+            case "WebrtcPeriodic": {
+                while ((System.currentTimeMillis() - canaryStartTime.getTime()) < canaryRunTime * 1000) {
+                    calculateTimeToFirstFragment();
+                }
+                System.exit(0);
```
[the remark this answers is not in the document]
stefankiesz: This wasn't terminating after the while loop, realized it is due to missing `executorService.shutdown();` after awaiting the futureTask which fixed the issue.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+
+            final Future<?> task = executorService.submit(getMediaWorker);
+            task.get();
+
+        } catch (Exception e) {
+            log.error(e);
+        }
+    }
+
+    protected static void publishMetricToCW(String metricName, double value, StandardUnit cwUnit) {
```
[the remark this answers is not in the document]
stefankiesz: So that the `RealTimeFrameProcessor` class which extends the `WebrtcStorageCanaryConsumer` class can call that same metric publishing function.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+                                                  // typical fragment duration coming from media server)
+
+                // NOTE: Metric publishing will NOT begin if canaryRunTime is < intervalInitialDelay
+                intervalMetricsTimer.scheduleAtFixedRate(intervalMetricsTask, intervalInitialDelay, intervalDelay);
+                Thread.sleep(canaryRunTime * 1000);
+                intervalMetricsTimer.cancel();
+                break;
+            }
+            case "WebrtcPeriodic": {
+                while ((System.currentTimeMillis() - canaryStartTime.getTime()) < canaryRunTime * 1000) {
```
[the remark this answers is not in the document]
stefankiesz: Hmm don't see how it could be an infinite loop, can you please explain?

We do only need to calculate it once, but we need to keep retrying until there are frames returned by `GetMedia()`. `GetMedia()` will return after ~3 sec if no footage is available. Once footage is available, it will continue to visit frames as long as there is footage available. Adding a comment to explain this.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+        } catch (Exception e) {
+            log.error(e);
+        }
+    }
+
+    public static void main(final String[] args) throws Exception {
+
+        // Import configurable parameters.
+        final Integer canaryRunTime = Integer.parseInt(System.getenv("CANARY_DURATION_IN_SECONDS"));
+        streamName = System.getenv("CANARY_STREAM_NAME");
```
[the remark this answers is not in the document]
stefankiesz: Do you mean for the constants file to contain a mapping of canary variables to env var names such as
```
CANARY_STREAM_NAME_ENV_VAR_NAME = "CANARY_STREAM_NAME";
```
and leave the parsing of env vars in main, or just have them be parsed and assigned in the constants file, or both?

canary/webrtc-c/src/media-server-storage/Common.cpp
```diff
@@ -0,0 +1,1864 @@
...
+        DLOGW("Could not get ICE agent metrics (0x%08x)", retStatus);
+    } else {
+        Canary::Cloudwatch::getInstance().monitoring.pushKvsIceAgentMetrics(&pSampleStreamingSession->iceMetrics);
+    }
+
+CleanUp:
+    return retStatus;
+}
+
+STATUS sendProfilingMetricsTryer(PSampleStreamingSession pSampleStreamingSession)
```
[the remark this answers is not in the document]
stefankiesz: Ok changed to Thread. I had called it a tryer (misspelled trier) due to it continuously trying to make a successful `sendProfilingMetrics` call in a while loop.

canary/webrtc-c/src/media-server-storage/Common.cpp
```diff
@@ -0,0 +1,1864 @@
...
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    PSampleConfiguration pSampleConfiguration;
+    
+    CHK(pSampleStreamingSession != NULL, STATUS_NULL_ARG);
+    pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
+    CHK(pSampleConfiguration != NULL, STATUS_NULL_ARG);
+
+    pSampleStreamingSession->pushProfilingThread = std::thread(sendProfilingMetricsTryer, pSampleStreamingSession);
+    pSampleStreamingSession->pushProfilingThread.detach();
```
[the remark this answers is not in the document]
stefankiesz: By detaching the thread, we don't have to wait on any processes to finish as we would with join(). This thread has a THREAD_SLEEP that `freeSampleStreamingSession()` would at the very minimum have to wait on, and although it is only a 100ms sleep, this would have a significant affect on the `storageDisconnectToFrameSentTime` metric.

canary/webrtc-c/src/media-server-storage/Common.cpp
stefankiesz: Do you prefer having two env vars for the whole relative paths (one for consumer, one for webrtc) or just have the env var handle the file name and leave the relative paths hardcoded?

canary/webrtc-c/src/media-server-storage/Common.cpp
stefankiesz: That's correct, this is just from the sample code.

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
stefankiesz: Joining here would cause us to wait here, but we need to get to the `sessionCleanupWait(...)` just below and let the `canaryDurationThread` exist in the background.

canary/webrtc-c/src/media-server-storage/Common.cpp
stefankiesz: For now, I added a definition to the WebRTC end:
`#define STORAGE_CANARY_FIRST_FRAME_TS_PATH      "../firstFrameSentTimeStamp.txt"
`

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,212 @@
...
+            Boolean newFragmentReceived = false;
+
+            final FutureTask<List<CanaryFragment>> futureTask = new FutureTask<>(
+                    new CanaryListFragmentWorker(mStreamName, mCredentialsProvider, listFragmentsEndpoint,
+                            Regions.fromName(mRegion), fragmentSelector));
+            Thread thread = new Thread(futureTask);
+            thread.start();
+
+            List<CanaryFragment> newFragmentList = futureTask.get();
+            if (newFragmentList.size() > fragmentList.getFragmentList().size()) {
```
sirknightj, line 84: Curious to know if it's possible that 2 fragments expired, but then 1 new fragment came in, this check wouldn't go through. It seems that we assume that the stream media expiration is long enough, but may want to leave a comment about this edge case

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
```diff
@@ -0,0 +1,111 @@
...
+
+                for (Fragment f : result.getFragments()) {
+                    fragments.add(new CanaryFragment(f));
+                }
+                nextToken = result.getNextToken();
+            }
+
+            // Fragments in the return of a listFragments call are in no particular order,
+            // so let's sort them by ascending fragment number
+            fragments.sort(Comparator.comparing(CanaryFragment::getFragmentNumberInt));
```
sirknightj, line 97: Wondering if the sorting is necessary. What is it used for?

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+                    .withAPIName(APIName.LIST_FRAGMENTS).withStreamName(streamName);
+            final String listFragmentsEndpoint = amazonKinesisVideo.getDataEndpoint(dataEndpointRequest)
+                    .getDataEndpoint();
+
+            TimestampRange timestampRange = new TimestampRange();
+            timestampRange.setStartTimestamp(canaryStartTime);
+            timestampRange.setEndTimestamp(new Date());
+
+            FragmentSelector fragmentSelector = new FragmentSelector();
+            fragmentSelector.setFragmentSelectorType("SERVER_TIMESTAMP");
```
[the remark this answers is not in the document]
sirknightj: Seems that you can use the enum value instead in this case: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/kinesisvideo/model/FragmentSelectorType.html

canary/webrtc-c/src/CloudwatchMonitoring.cpp
```diff
@@ -304,7 +330,9 @@ VOID CloudwatchMonitoring::pushKvsIceAgentMetrics(PKvsIceAgentMetrics pKvsIceAge
 VOID CloudwatchMonitoring::pushSignalingClientMetrics(PSignalingClientMetrics pSignalingClientMetrics)
 {
     MetricDatum offerToAnswerDatum, getTokenDatum, describeDatum, createDatum, endpointDatum,
-                iceConfigDatum, connectDatum ,createClientDatum, fetchDatum, connectClientDatum;
+                iceConfigDatum, connectDatum ,createClientDatum, fetchDatum, connectClientDatum, joinSessionToOfferDatum;
```
sirknightj, line 333:
```suggestion
                iceConfigDatum, connectDatum, createClientDatum, fetchDatum, connectClientDatum, joinSessionToOfferDatum;
```

canary/webrtc-c/src/CloudwatchMonitoring.cpp
```diff
@@ -355,6 +383,21 @@ VOID CloudwatchMonitoring::pushSignalingClientMetrics(PSignalingClientMetrics pS
     connectClientDatum.SetValue(pSignalingClientMetrics->signalingClientStats.connectClientTime);
     connectClientDatum.SetUnit(Aws::CloudWatch::Model::StandardUnit::Milliseconds);
     this->push(connectClientDatum);
+
+    if(this->isStorage) {
```
sirknightj, line 387:
```suggestion
    if (this->isStorage) {
```

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
stefankiesz: Not currently used for anything, but could be used to validate fragment order as media is ingested and to identify specific missing fragments in continuity. Will re-evaluate the necessity.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryFragment.java
```diff
@@ -0,0 +1,31 @@
...
+    public CanaryFragment(Fragment fragment) {
+        this.mFragment = fragment;
+    }
+
+    public Fragment getFragment(){
+        return this.mFragment;
+    }
+
+    public BigInteger getFragmentNumberInt() {
+        return new BigInteger(this.mFragment.getFragmentNumber());
```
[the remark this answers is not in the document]
stefankiesz: Can sort using string, but if using sorted fragments to confirm "monotonically increasing" nature of fragment numbers or any other validation that requires arithmetic, having them as numbers would be nice. Will re-evaluate the necessity of this.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/RealTimeFrameProcessor.java
```diff
@@ -0,0 +1,73 @@
...
+
+                // TODO: Add this to contants file for now
+                String filePath = "../webrtc-c/firstFrameSentTimeStamp.txt";
+                final String frameSentTimeStr = FileUtils.readFileToString(new File(filePath)).trim();
+                long frameSentTime = Long.parseLong(frameSentTimeStr);
+
+                // Check for a late consumer end start that may add to the measured
+                // FirstFrameSentToFirstFrameConsumed time
+                if (canaryStartTime > frameSentTime) {
+                    // Consumer started after master sent out first frame, not pushing the
```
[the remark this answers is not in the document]
stefankiesz: Log added

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
stefankiesz: Yes, good point, adding a comment.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryListFragmentWorker.java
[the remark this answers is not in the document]
disa6302: It can return empty list. Returning from `finally` block is not intuitive. `finally` is a block that is used to run cleanup or an operation that should be executed irrespective of exception. It is generally advisable to have return at the end of the function and not within these blocks. Parser library is a standalone sample and  although we are taking pieces of code from those samples to set this up, we should ensure this code follows good practices. We can look into cleaning those up wherever relevant in the future.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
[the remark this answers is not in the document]
disa6302: Lets add it anyways for completion purpose.

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
disa6302: We dont join here. We join after `sessionCleanUpWait()`, it should not hang at that point, correct?

canary/webrtc-c/src/media-server-storage/Common.cpp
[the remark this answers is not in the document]
disa6302: Why dont we start this thread after the session is set up? The metrics are all post start up right? Is there a reason we need to rush publishing this instead of waiting on P2P set up completion?

canary/webrtc-c/src/Include.h
```diff
@@ -97,6 +98,7 @@
 #define KVS_METRICS_INVOCATION_PERIOD        (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
 #define CANARY_METADATA_SIZE                 (SIZEOF(UINT64) + SIZEOF(UINT32) + SIZEOF(UINT32))
 
+#define STORAGE_CANARY_FIRST_FRAME_TS_PATH      "../firstFrameSentTimeStamp.txt"
```
disa6302, line 101: The file name should be an env. The same file name is being used in consumer and it is very easy to miss changing the name on both ends. An env solves this since it will be commonly used by consumer and webrtc.

Something like: `#define STORAGE_FIRST_FRAME_TS_FILE_PATH_ENV STORAGE_FIRST_FRAME_TS_FILE_PATH`

Usage would be: `export STORAGE_FIRST_FRAME_TS_FILE_PATH=<path-to-file>`

canary/webrtc-c/src/media-server-storage/Common.cpp
```diff
@@ -0,0 +1,1863 @@
...
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PSampleConfiguration pSampleConfiguration;
+    UINT32 i;
+    UINT64 data;
+    StackQueueIterator iterator;
+    BOOL locked = FALSE;
+
+    std::string filePath = "../firstFrameSentTimeStamp.txt";
```
disa6302, line 1327: Change this to use the env value. You can have this set up in `streamingSession` struct.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/RealTimeFrameProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void process(Frame frame, MkvTrackMetadata trackMetadata, Optional<FragmentMetadata> fragmentMetadata,
+            Optional<FragmentMetadataVisitor.MkvTagProcessor> tagProcessor) throws FrameProcessException {
+        if (!isFirstFrame) {
+            try {
+                long currentTime = System.currentTimeMillis();
+                long canaryStartTime = super.mCanaryStartTime.getTime();
+
+                // TODO: Add this to contants file for now
+                String filePath = "../webrtc-c/firstFrameSentTimeStamp.txt";
```
disa6302, line 41: Lets use the env reading here. Details in C part of the PR. Also, it is advisable to create this file in the `pwd` (in our case the build directory). After the run completes, or after the file is read, you can just delete the file in Java code, or if not programatically, removeFile in Jenkins groovy script.

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,215 @@
...
+                .withRegion(mRegion)
+                .withCredentials(mCredentialsProvider)
+                .build();
+        mCwClient = AmazonCloudWatchClientBuilder.standard()
+                .withRegion(mRegion)
+                .withCredentials(mCredentialsProvider)
+                .build();
+
+        switch (mCanaryLabel) {
+            case "WebrtcLongRunning": {
```
disa6302, line 177: All these should be defined in a Constants.java class and used. It is not a good programming practice to hard code strings, because it makes it cumbersome to update all the locations it could be used in. Something like:

```
public final class Constants {

    // Example constants
    public static final String WEBRTC_LONG_RUNNING_LABEL = "WebrtcLongRunning";
    public static final STRING INTERVAL_TIMER_NAME = "IntervalMetricsTimer";

    // More constants can be added here

    private Constants() {
    }
}
```
  When you use it here, it should be:
  
  ```
  case Constants.WEBRTC_LONG_RUNNING_LABEL:
  ```

disa6302 (changes requested):

canary/webrtc-c/src/media-server-storage/Common.cpp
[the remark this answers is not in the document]
stefankiesz: Ah yes, no need for metric publishing to be initiated here. I moved it to be out of sync with the general flow in the `onConnectionStateChange` callback and switched to using `join`.

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
stefankiesz: We remain within `sessionCleanUpWait` in a while loop for the remainder of the run though, so we wouldn't ever reach that code. I moved the this code over into `sessionCleanUpWait` just above the while loop, but still need to use detach. We could do away with the whole shutdown thread and just use that while loop to check for duration reached, how does that sound?

stefankiesz:
The WebRTC PR for easy diff viewing is updated too: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1847

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/CanaryConstants.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static final String PERIODIC_LABEL = "WebrtcStoragePeriodic"; // Short period, used for time to first frame consumed metrics.
+    public static final String SUB_RECONNECT_LABEL = "WebrtcStorageSubReconnect"; // < 1 Media Server forced reconnect, currently must be 60min.
+    public static final String SINGLE_RECONNECT_LABEL = "WebrtcStorageSingleReconnect"; // 1 Media Server forced reconnect, currently must be 60-120min.
+    public static final String EXTENDED_LABEL = "WebrtcStorageExtended"; // > 1 Media Server forced reconnect, currently must be > 120min.
+
+    public static final String CW_DIMENSION_INDIVIDUAL = "StorageWebRTCSDKCanaryStreamName";
+    public static final String CW_DIMENSION_AGGREGATE = "StorageWebRTCSDKCanaryLabel";
+
+    public static final String FIRST_FRAME_TS_FILE_PATH = "../webrtc-c/build/";
```
disa6302, line 29: We can probably keep this at root instead of build given that is the common directory for both consumer and webrtc

canary/webrtc-c/src/media-server-storage/kvsWebRTCClientMaster.cpp
disa6302: Yes. Lets do that. No need of a new thread if `sessionCleanUpWait` is going to run through and through anyways.

disa6302 (approved):

canary/consumer-java/src/main/java/com/amazon/kinesis/video/canary/consumer/WebrtcStorageCanaryConsumer.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    private static void calculateTimeToFirstFragment() {
+        try {
+            final StartSelector startSelector = new StartSelector()
+                    .withStartSelectorType(StartSelectorType.PRODUCER_TIMESTAMP).withStartTimestamp(mCanaryStartTime);
+
+            final long currentTime = new Date().getTime();
+            double timeToFirstFragment = Double.MAX_VALUE;
```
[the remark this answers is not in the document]
stefankiesz: Fixed

disa6302 (approved):

sirknightj (approved):

PR #339, opened 2024-01-23
Shorten storage canary labeles

stefankiesz:
The "Webrtc" part of the label is redundant as the channel and stream names begin with "webrtc," so removed that portion of the labels to make channel/stream names less long.


disa6302:
Can we make the same change for the rest to be consistent?

disa6302 (approved):

PR #367, opened 2024-07-02
Wc 1.10.3

disa6302:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/canary/webrtc-c/src/Peer.cpp
+++ b/canary/webrtc-c/src/Peer.cpp
@@ -379,9 +379,10 @@ STATUS Peer::connect()
         SignalingMessage msg;
 
         MEMSET(&offerSDPInit, 0, SIZEOF(offerSDPInit));
-        CHK_STATUS(createOffer(this->pPeerConnection, &offerSDPInit));
+        DLOGI("Trickle ICE: %s", this->trickleIce ? "Enabled" : "Disabled");
+        offerSDPInit.useTrickleIce = this->trickleIce;
         CHK_STATUS(setLocalDescription(this->pPeerConnection, &offerSDPInit));
-
+        CHK_STATUS(createOffer(this->pPeerConnection, &offerSDPInit));
         if (!this->trickleIce) {
             CHK_STATUS(this->awaitIceGathering(&offerSDPInit));
         }
--- a/canary/webrtc-c/src/Peer.cpp
+++ b/canary/webrtc-c/src/Peer.cpp
@@ -191,8 +191,35 @@ CleanUp:
 
 STATUS Peer::initRtcConfiguration(const Canary::PConfig pConfig)
 {
+    auto awaitGetIceConfigInfoCount = [](SIGNALING_CLIENT_HANDLE signalingClientHandle, PUINT32 pIceConfigInfoCount) -> STATUS {
+        STATUS retStatus = STATUS_SUCCESS;
+        UINT64 elapsed = 0;
+
+        CHK(IS_VALID_SIGNALING_CLIENT_HANDLE(signalingClientHandle) && pIceConfigInfoCount != NULL, STATUS_NULL_ARG);
+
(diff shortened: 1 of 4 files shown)


stefankiesz (approved):

canary/webrtc-c/src/Peer.cpp
```diff
@@ -214,13 +241,31 @@ STATUS Peer::initRtcConfiguration(const Canary::PConfig pConfig)
...
+        // Set the URIs from the configuration
+        CHK_STATUS(awaitGetIceConfigInfoCount(signalingClientHandle, &iceConfigCount));
         /* signalingClientGetIceConfigInfoCount can return more than one turn server. Use only one to optimize
          * candidate gathering latency. But user can also choose to use more than 1 turn server. */
         for (uriCount = 0, i = 0; i < MAX_TURN_SERVERS; i++) {
             CHK_STATUS(signalingClientGetIceConfigInfo(signalingClientHandle, i, &pIceConfigInfo));
-            CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
-            uriCount += pIceConfigInfo->uriCount;
+            for (j = 0; j < pIceConfigInfo->uriCount; j++) {
+                CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
```
sirknightj, line 251: Not sure where CHECK is defined but this seems redundant if max count is always 1. Are we planning on adding the different turn server Urls (turn with tcp, or turns w/ udp)

canary/webrtc-c/src/Peer.cpp
```diff
@@ -335,9 +379,10 @@ STATUS Peer::connect()
         SignalingMessage msg;
 
         MEMSET(&offerSDPInit, 0, SIZEOF(offerSDPInit));
-        CHK_STATUS(createOffer(this->pPeerConnection, &offerSDPInit));
+        DLOGI("Trickle ICE: %s", this->trickleIce ? "Enabled" : "Disabled");
+        offerSDPInit.useTrickleIce = this->trickleIce;
         CHK_STATUS(setLocalDescription(this->pPeerConnection, &offerSDPInit));
-
+        CHK_STATUS(createOffer(this->pPeerConnection, &offerSDPInit));
```
sirknightj, line 385: Why are we creating the offer after setting local description?

canary/webrtc-c/src/Peer.cpp
stefankiesz: CHECK is defined in PIC PlatformUtils. Not planning on adding more turn server urls, but leaving in case we ever do and to replicate the same way we do this in the samples [here](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/30f52d18e647747de1ec29ad6f6d3f62128f5494/samples/Common.c#L383) with maxTurnServer set to 1 [here](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/30f52d18e647747de1ec29ad6f6d3f62128f5494/samples/Common.c#L353).

canary/webrtc-c/src/Peer.cpp
stefankiesz: This is also done in the sample. It seems that it is done so to kick off `iceAgentStartGathering` as soon as possible which is done via `setLocalDescription`, and `setLocalDescription` does not use `&offerSDPInit` other than to verify that it is not null.

sirknightj: In the JS at least, it is like this. Seeing it the other way around is really odd.
```
console.log('[VIEWER] Creating SDP offer');
await viewer.peerConnection.setLocalDescription(
    await viewer.peerConnection.createOffer({
        offerToReceiveAudio: true,
        offerToReceiveVideo: true,
    }),
);
```

sirknightj (approved):

PR #409, opened 2025-04-29
Handle linked resources for the webrtc ingestion feature

sirknightj:
*Issue #, if available:*
- N/A

*Description of changes:*
- Handle the case for linked resources. Currently, if the resources are linked, the following error occurs.

```
An error occurred (ResourceInUseException) when calling the DeleteStream operation: Disable MediaStorageConfiguration to delete the stream
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ResourceInUseException: An error occurred (ResourceInUseException) when calling the DeleteStream operation: Disable MediaStorageConfiguration to delete the stream
```
- The script will call delete, and if the error occurs, will first unlink and then try deleting again. The alternative (not implemented) would be to call describe_mapped_resource_configuration/describe_media_storage_configuration for every single resource before calling delete, which is not efficient since it will double the API calls. Considering the majority use case is not having resources linked/configured for ingestion, this seems like a great trade-off.
- The script is open-source and easy to change, so anyone can change it however they want.

### DIFF
--- a/python/cleanup/kvs_cleanup.py
+++ b/python/cleanup/kvs_cleanup.py
@@ -164,14 +164,27 @@ def delete_old_test_channels(channel_regex: str, dry_run: bool, time_threshold):
 
 def unlink_resources(stream_arn: str = None, channel_arn: str = None):
     if not stream_arn and not channel_arn:
-        raise SyntaxError('Please provide either a stream arn or channel arn')
+        raise ValueError('Please provide either a stream arn or channel arn')
 
     if not channel_arn:
         response = client.describe_mapped_resource_configuration(StreamARN=stream_arn)
         channel_arn = response.get('MappedResourceConfigurationList')[0].get('ARN')
 
     print(f'Unlinking: {stream_arn=}, {channel_arn=}')
-    client.update_media_storage_configuration(ChannelARN=channel_arn, MediaStorageConfiguration={'Status': 'DISABLED'})
+    for attempt in range(5):
+        try:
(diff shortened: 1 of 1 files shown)


python/cleanup/kvs_cleanup.py
```diff
@@ -145,6 +162,18 @@ def delete_old_test_channels(channel_regex: str, dry_run: bool, time_threshold):
     return channel_count, channels_deleted
 
 
+def unlink_resources(stream_arn: str = None, channel_arn: str = None):
+    if not stream_arn and not channel_arn:
+        raise SyntaxError('Please provide either a stream arn or channel arn')
```
unicornss, line 167: ValueError

python/cleanup/kvs_cleanup.py
```diff
@@ -145,6 +162,18 @@ def delete_old_test_channels(channel_regex: str, dry_run: bool, time_threshold):
...
+
+    if not channel_arn:
+        response = client.describe_mapped_resource_configuration(StreamARN=stream_arn)
```
unicornss, line 170: try except?

python/cleanup/kvs_cleanup.py
sirknightj: Adjusted

python/cleanup/kvs_cleanup.py
sirknightj: Added

unicornss (approved):

